Issue #5, opened 2023-03-04
Set up ReadTheDocs


jtpio:
### Problem

Follow-up to https://github.com/jupyterlite/pyodide-kernel/pull/2.

We should set up ReadTheDocs be able to test changes easily on each PR.

### Suggested Improvement

Set up ReadTheDocs with `jupyterlite-sphinx` like https://xeus-python-kernel.readthedocs.io/en/latest/.

jtpio:
Closing as fixed by #8

PR #6, opened 2023-03-04
Add files for ReadTheDocs

jtpio:
Fixes https://github.com/jupyterlite/pyodide-kernel/issues/5

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,7 +32,7 @@ def on_config_inited(*args):
 
     HERE = Path(__file__)
     ROOT = HERE.parent.parent
-    subprocess.check_call([sys.executable, "-m", "pip", "install", "'.[dev]'"], cwd=str(ROOT))
+    subprocess.check_call([sys.executable, "-m", "pip", "install", "\".[dev]\""], cwd=str(ROOT))
     subprocess.check_call(["yarn"], cwd=str(ROOT))
     subprocess.check_call(["yarn", "build"], cwd=str(ROOT))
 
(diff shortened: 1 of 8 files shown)


docs/conf.py
```diff
@@ -0,0 +1,41 @@
...
+# }
+
+def on_config_inited(*args):
+    import sys
+    import subprocess
+    from pathlib import Path
+
+    HERE = Path(__file__)
+    ROOT = HERE.parent.parent
+    subprocess.check_call([sys.executable, "-m", "pip", "install", "\".[dev]\""], cwd=str(ROOT))
```
bollwyvl, line 35: i've found that doing this in the `jobs` chunk of `.readthedocs.yml` helps a lot in debugging, as errors inside the RTD/sphinx output are... very hard to read.

bollwyvl: Indeed, it can even replace many of the needs for the conda junk with:

```
version: 2

build:
  os: ubuntu-22.04
  tools:
    python: "3.10"
    nodejs: "18"

python:
  install:
    - method: pip
      path: .
      extra_requirements:
        - docs
        - lite
```

thinking about it more: when building the docs locally, should probably _not_ force a re-install of the dev enviornment (which in turn will spawn _more_ environments, due to the build backend), and a hard js rebuild, if fixing one little typo.

jtpio: Thanks for the hint, will try it out.

bollwyvl: didn't want to step on your toes, so #8.

jtpio:
Closing as fixed by #8

PR #7, opened 2023-03-04
Remove Binder files

jtpio:
Simplify the repo by removing the Binder files since they won't be used.


bollwyvl:
Hmm... _removing_ a previously-working binder config makes me sad: there's a lot to be said for having a reproducible environment where PR reviews can be taken for a spin, etc. even if it's not a focus of the product being built.

Further, undoubtedly this repo will be emulated/copy-pasted in downstreams, and having a single-file way to get up to a working environment is pretty clutch.

PR #8, opened 2023-03-04
Add files for ReadTheDocs, enable Python tests, maintenance

### LINKED ISSUE #9: Enable Python tests for the addon
### Problem

Follow-up to #2

Currently the Python tests have been disabled: https://github.com/jupyterlite/pyodide-kernel/blob/9d3ee654db277011b0dea9280aa1ec192ffd4e8c/.github/workflows/test.yml#L39-L42

Because the released `jupyterlite` provides the same addons and creates conflicts.

### Proposed Solution

Re-enable these tests. 

### Additional context

This might have to happen a bit later after completing https://github.com/jupyterlite/jupyterlite/pull/854.

bollwyvl:
## References

- extends/includes #6
- fixes #9

## Changes

- [x] move docs deps to `pyproject.toml`
- [x] use more `.readthedocs.yml` features for `apt`, `pip`, etc.
- [x] get more single-source-of-truth data from 
  - [x] `pyproject.toml` (also update URLs, etc)
  - [x] `jupyterlite_pyodide_kernel.__verrsion__`
- [x] consume the built jupyterlite wheel from https://github.com/jupyterlite/jupyterlite/pull/854 
- [x] add a `jlpm quickstart`
- [x] slightly improve the `nx` config to ensure the wheels get built
  - > it's very worrisome that it doesn't fail if something is missing, but at least it seems to not cache as poorly
- [x] drop `jupyterlite-sphinx` for now
  - > while useful for "general purpose" sites, it seemed to have some issues with some of the config and paths needed for extensions at the level this one has to be
- [x] fix detection and bookkeeping for federated `wheelDir`s
  - > because now that's the only way any wheels will ever show up)
- [x] add `package.json#/piplite/wheelDir` to actually find the files
- [x] restore `pytest`
- [x] format more things with `prettier`


bollwyvl:
:crying_cat_face: 
```
Error

Concurrency limit reached (2), retrying in 5 minutes.
```

bollwyvl:
@jtpio this is at least drawing a site with ipywidgets, and would unblock downstream work.

CONTRIBUTING.md
```diff
@@ -0,0 +1,98 @@
...
+- `nodejs >=18,<19`
+
+## Quick Start
+
+The command below cleans out all built artifacts, ensures a fully-configured JS/Python
+environment, and generates everything up to the documentation website and test reports.
+
+```bash
+git clone https://github.com/jupyterlite/pyodide-kernel
+cd pyodide kernel
```
jtpio, line 16:
```suggestion
cd pyodide-kernel
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,98 @@
...
+npm run quickstart
+```
+
+> The above installs JupyterLab, which provides the `jlpm` command, a pre-packaged
+> version of `yarn`, and is used instead of `npm` from here on out.
+
+### Serve
```
jtpio, line 23: It would be nice to keep a small section about the dev workflow like the original readme / cookiecutter does. Basically what sequence of commands should be run and when to expect changes to be effective (for example after re-running `jlpm run build`).

CONTRIBUTING.md
```diff
@@ -0,0 +1,98 @@
...
+
+```bash
+jlpm serve
+```
+
+> Serve the `build/` directory as `http://127.0.0.1:8000`, which contains:
```
jtpio, line 29: Maybe we can drop the quotes (`>`) for better readability?

jupyterlite_pyodide_kernel/tests/conftest.py
```diff
@@ -6,6 +8,60 @@
...
+
+
+@pytest.fixture
+def index_cmd():
+    """get the command line arguments for indexing a folder."""
+    return ("jupyter", "piplite", "index")
+
+
+@pytest.fixture
+def a_pydodide_tarball():
```
jtpio, line 31:
```suggestion
def a_pyodide_tarball():
```

jtpio:
> * it seemed to have some issues with some of the config and paths needed for extensions at the level this one has to be

What kind of issues exactly? Wondering if they should be reported upstream.

bollwyvl:
Thanks for the review: in away from the keys at the moment: feel free to
fix anything directly that's obvious.

CONTRIBUTING.md
bollwyvl: > a small section

I've expanded it some, replicating some of the quick start section, with deeper links to invidividual pieces.

As the "front door" of a package for end users (e.g. shows up on PyPI), I think having too much nodejs-related stuff is actually a shortcoming of the existing cookiecutter output, partially as a hangover from the bad old days of source extensions.

CONTRIBUTING.md
bollwyvl: 8f9245a , added table

jupyterlite_pyodide_kernel/tests/conftest.py
bollwyvl: 8f9245a

bollwyvl:
> What kind of issues exactly? 

Just the plain old, "binary (nodejs) packages doesn't work very well on everyone's machines". lerna depends on the nwrl stack, which has some rust-based components (with C depenedencies) that doesn't always install correctly. I've run into similar issues with `esbuild`.

Like in python for jupyterlab 3.6+, the adoption of more esoteric build chains (and i though `node-gyp` was bad) have created situations so a first time user can't even _get_ to starting their workflow, and are seeing deep, platform-specific issues.

It's not really worth my time to go fighting upstream... if we can keep CI running, it's fine for me, but I don't look forward to supporting everyone's special snowflake machine (see the JupyterLab upstream issue about `canvas` on windows and apple silicon).

CONTRIBUTING.md
bollwyvl: 8f9245

jtpio (approved):
Thanks :+1:

PR #13, opened 2023-03-06
Update Jupyter Releaser hooks

jtpio:
This repo now differs from the extension cookiecutter and does not rely on `hatch-jupyter-builder` which seems to be leading to some assets not being properly included.

- [ ] Replicate `ensured-targets` provided by  `tool.hatch.build.hooks.jupyter-builder`


jtpio:
OK the wheel built by the releaser now looks better, trying locally:

image


> [ ] Replicate `ensured-targets` provided by `tool.hatch.build.hooks.jupyter-builder`

Will track that separately to not block the first release.

PR #16, opened 2023-03-06
Releaser fixes

jtpio:
Follow-up to a failed release: https://github.com/jupyterlite/pyodide-kernel/actions/runs/4344479404/jobs/7587852938

- [x] Revert publish commit that bumped to `0.0.2`
- [x] Add `CHANGELOG.md` to `.prettierignore` to avoid CI failure after the release commit (changelog is already formatter by the releaser)
- [ ] Fix direct dependency preventing the upload to PyPI


bollwyvl:
Before doing an actual (even pre-) release, we should probably put `jupyterlite` back in the main deps segment, and move the icky RTD link to `[dev]` or something...

jtpio:
> we should probably put `jupyterlite` back in the main deps segment, and move the icky RTD link to `[dev]` or something...

Yeah this is probably enough for now indeed, so we can unblock https://github.com/jupyterlite/jupyterlite/pull/854.

bollwyvl (approved):
LGTM

jtpio:
Alright so even in `dev` we face the same issue:

image

Maybe we can just manually install from the direct link on CI and mention that in the contributing docs for now.

jtpio:
Will follow-up with another PR.

PR #19, opened 2023-03-06
Restore more python tools and tests

### LINKED ISSUE #18: Restore more of the python tools and tests
### Problem

Breaking off from the upstream repo, we lost:

- `black`
- `pyflakes`
- tests of many of the CLI pieces
  - `check` (and control of the `piplite` schema)

### Proposed Solution

- [ ] add back some more tests
- [ ] add some more python tools under `jlpm lint`

### Additional context

...

bollwyvl:
## References

- fixes #18

## Changes

- [x] restore `black`, `isort`, `pyflakes`
- [x] add above to `jlpm lint` 
- [x] restore `check` tests
- [x] maybe add `status`
- [x] updates wheel reference to new URL
  - >  `https://jupyterlite.rtfd.io/en/latest/_static/jupyterlite-0.1.0b18-py3-none-any.whl` 
- [x] figure out `piplite` schema story
  - [x] copy schema into `packages/pyodide-kernel/schema`
  - [x] hoist with webpack
  - [x] use in `check`
- [x] update `outputDir` to point at accurate `npm` name
- [x] rename some constants from `PYOLITE_*` to `PYODIDE_*`
- [x] don't use `pyodide-core` in demo
  - ~~add fallback micropip url~~
    - the `pyodide-core` distribution does not include `micropip `
    - it's too early for it to be found in `pipliteUrls`
  - would prefer to solve this properly with #14 than some in-between measure 
- [x] figure out what to do with `litePluginSettings`
  - [x] hoist fragment of upstream settings
  - [x] use in `check`
  - [x] add `jupyter-lite.ipynb`s to `check` (maybe more)
  - [x] upstream PR to remove old settings https://github.com/jupyterlite/jupyterlite/pull/998
- [x] add an `intro.ipynb`
- [x] use the upstream pyodide logo
  - > legit for now (other than dark mode): working this upstream on https://github.com/pyodide/pyodide/issues/3634  
- [ ] ???

## Open Questions
- [x] Do we want to do a bridge for the old `litePluginSetings/@jupyterlite/pyolite-kernel`? 
  -  **No**
    - > Similar to the switch to PipliteAddon I would say it's fine without, and we document the new name in the changelog. It is also searchable on GitHub and we can track the number of uses.
  - this is no observable impact on `jupyter_lite_config.json`-driven stuff
    - but heavily tweaked `jupyter-lite.json` _would_
  - we _didn't_ before when we hoisted `PyodideAddon` and `PipliteAddon`


jtpio:
How about adopting `ruff` for some of these linting tasks?

It's been used in other Jupyter projects like JupyterLab, Notebook, Jupyter Server, nbclient, nbformat. And it would help reduce the number of different tools used here.

bollwyvl:
Sure, we can go with `ruff` instead of pyflakes and `isort`, but it can't replace `black` yet, no?

bollwyvl:
Ah, finally got to the reproducer of `check`:

```
Task generation 'check' has duplicated definition of 'check:jupyterlite-pyodide-kernel-piplite:validate:./extensions/@jupyterlite/pyodide-kernel-extension/static/pypi/all.json'
```

Looks like we're not deduplicating the `pipliteWheelUrls`.

jtpio:
> Sure, we can go with `ruff` instead of pyflakes and `isort`, but it can't replace `black` yet, no?

Right: https://beta.ruff.rs/docs/faq/#is-ruff-compatible-with-black

> Yes. Ruff is compatible with [Black](https://github.com/psf/black) out-of-the-box, as long as the line-length setting is consistent between the two.

bollwyvl:
Great. I've turned on `ruff` with the defaults (one skip for our nasty patch-laden imports in the kernel itself).

As a follow on we can turn on as many rules as we want from its vast stable. However, for example, I'd much rather have mypy doing typechecking rather than relying on the ruff semi-implementation, which just complains for complaining sake, and doesn't always agree with mypy.

jtpio:
> Do we want to do a bridge for the old `litePluginSetings/@jupyterlite/pyolite-kernel`?

Similar to the switch to `PipliteAddon` I would say it's fine without, and we document the new name in the changelog. It is also searchable on GitHub and we can track the number of uses.

packages/pyodide-kernel-extension/schema/plugin.v0.schema.json
```diff
@@ -0,0 +1,63 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://jupyterlite-pyodide-kernel.readthedocs.org/en/latest/reference/schema/settings-v0.html#",
+  "title": "Pyodide Kernel Settings Schema v0",
+  "description": "a schema for user-serviceable customizaton of a JupyterLite sPyodide Kernel",
```
jtpio, line 5:
```suggestion
  "description": "a schema for user-serviceable customization of a JupyterLite Pyodide Kernel",
```

jtpio:
> Similar to the switch to `PipliteAddon` I would say it's fine without, and we document the new name in the changelog. It is also searchable on GitHub and we can track the number of uses.

Also last time I used the GitHub search to find repos using the `piplite_urls` and opened PRs to help them migrate to `PipliteAddon`. Planning to do a similar check after the next JupyterLite release as well.

jtpio:
Thanks @bollwyvl for this.

It's looking good, happy to merge and iterate if it's ready :+1:

pyproject.toml
```diff
@@ -147,3 +154,8 @@ before-build-python = [
 
 [tool.check-wheel-contents]
 ignore = ["W002"]
+
+[tool.isort]
```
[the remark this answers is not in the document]
bollwyvl: nope! 7619a47

packages/pyodide-kernel-extension/schema/plugin.v0.schema.json
bollwyvl: fixed, and renamed, (by other means) on c9b7d64

bollwyvl:
>  iterate if it's ready

Yeah, there are a few remaining uncovered things that would be nice (e.g. testing the normalizing `PythonPackage` to `pythonpackage` in indexes, actually quiet a headache!). But gotta stop somewhere!

It's probably fine to keep the newer addon base bits here for a tick, just to kick the tires a bit before moving them upstream, and shouldn't block this PR anyway.

bollwyvl:
RTD seems happy.

One little nit I'm working: I think we want at least _something_ in the `examples` folder, and maybe open it in `retro` for the iframe.

I'll try to repurpose some content from upstream.

jtpio:
> I'll try to repurpose some content from upstream.

This is still an open question following up https://github.com/jupyterlite/jupyterlite/pull/854.

For now it's nice to have the Pyodide examples available on the main site: https://jupyterlite.rtfd.io/en/latest/try/lab.

But at some point we might indeed want to let each kernel repo handle their examples. And feature them in https://github.com/jupyterlite/demo which uses released versions.

bollwyvl:
Right: i mean _like anything at all_ (e.g. `intro.ipynb`)

packages/pyodide-kernel-extension/src/index.ts
```diff
@@ -13,6 +13,7 @@ import { IKernel, IKernelSpecs } from '@jupyterlite/kernel';
 import { IBroadcastChannelWrapper } from '@jupyterlite/contents';
 
 export * as KERNEL_SETTINGS_SCHEMA from '../schema/kernel.v0.schema.json';
+import * as KERNEL_ICON from '!!url-loader!../style/img/pyodide.svg';
```
jtpio, line 16: Normally we should be able to do without `url-loader` and use the webpack 5 Asset Module instead.

jtpio: For example the p5 kernel brings its own logo here: https://github.com/jupyterlite/p5-kernel/blob/fbda9a78adde48e7ed9012886c28f68c81e9270d/packages/p5-kernel-extension/src/index.ts#L15

Although it's a png.

bollwyvl: sure, asset loader is rad.

bollwyvl:
Gah, webpack. All of the configurations i tried ended up mangling the SVG in some way to where it was broken, usual encoded as a string with `export default` yadda yadda up front.

Reliably getting the string worked (a la labicons), so just moving forward with a base64-encoded data uri.

bollwyvl:
Also optimized the SVG (did it offline: no `svgo` plz), and added a white background (but no border)  to the pyodide logo for now:

| light | dark |
|-|-|
| image | Screenshot from 2023-03-09 16-57-16

packages/pyodide-kernel-extension/src/index.ts
bollwyvl: it's "working" e0457fa

bollwyvl:
RTD looking fine:

Screenshot from 2023-03-09 17-09-43

jtpio (approved):
Thanks!

PR #22, opened 2023-03-09
Depend on `jupyterlite-core`

jtpio:
Builds on top of #19.

Testing the Pyodide kernel against https://github.com/jupyterlite/jupyterlite/pull/994.


bollwyvl:
I think to avoid duplication/churn while working this arc, we probably need a `requirements-upstream.txt` for now that captures the icky RTD wheels deps explicitly in one place.

PR #23, opened 2023-03-10
Depend on `jupyterlite-core`

jtpio:
Follow-up to https://github.com/jupyterlite/jupyterlite/pull/994.

Planning to make a first pre-release for `jupyterlite-core` from `jupyterlite/jupyterlite` so we can start using it here.


jtpio:
Yes, and CI is also passing.

bollwyvl (approved):
:guitar:

PR #24, opened 2023-03-10
Update `jupyterlite-core`

jtpio:
Follow-up to https://github.com/jupyterlite/jupyterlite/pull/998


jupyterlite_pyodide_kernel/addons/_base.py
```diff
@@ -1,94 +0,0 @@
-"""common addon features for ``jupyterlite-pyodide-kernel``
```
bollwyvl, line 1: was intending on using this to extract more common things from the two plugins, aside from `pyodide_settings`: `piplite` will need to be more aware of `pyodide` stuff for #14.

jtpio: Sure. I can put the PR back to draft and let you pick it up if you want.

bollwyvl: :blush: i'll try and work on porting https://github.com/jupyterlite/jupyterlite/pull/979 and see what turns up

jtpio:
More happening in https://github.com/jupyterlite/pyodide-kernel/pull/30 and #27.

PR #26, opened 2023-03-10
Releaser: support bumping versions with `next` as the spec

jtpio:
This will make is easier to make incremental releases.

Add similar logic as in `jupyter-releaser`: https://github.com/jupyter-server/jupyter_releaser/blob/cecae82d753f2ac3b741c05ddf5302d64ecfe3a7/jupyter_releaser/util.py#L246-L348

But keeping things simpler (just `hatch` and `next`).


jtpio:
Releaser assets look good:

image

Issue #37, opened 2023-03-30
Use the `comm` package


jtpio:
### Problem

It should now be possible to use the new `comm` package: https://pypi.org/project/comm/

Currently there is a mock for `ipykernel` here: https://github.com/jupyterlite/pyodide-kernel/tree/main/packages/pyodide-kernel/py/ipykernel

Which hopefully could be removed.

### Proposed Solution

Use the `comm` package so can get rid of one of the mocks.

### Additional context

https://github.com/jupyterlite/jupyterlite/issues/18

jtpio:
Actually not sure since `ipywidgets` initially dropped `ipykernel`, but then added it back in https://github.com/jupyter-widgets/ipywidgets/pull/3749.

jtpio:
Hmm looks like the latest `ipywidgets` releases break widgets in the Pyodide kernel: 

![image](https://user-images.githubusercontent.com/591645/228942077-7c660c1d-debd-466c-9a1d-82b970851621.png)

The latest that seems to work is `8.0.4`: 

![image](https://user-images.githubusercontent.com/591645/228942210-13ce8a64-5815-4070-942b-ccc25dc22390.png)

jtpio:
> The latest that seems to work is `8.0.4`:

cc @martinRenou who might know if this is related to a change in `ipywidgets` in `8.0.5` or `8.0.6`.

bollwyvl:
Yep bottom pins are tighter now, basically to get around the
structuredClone SNAFU. Need to bump the mock packages... Older 8.x and 7.x
ones will be compatible.

But also should be pinning for reproducible builds with front and backend.

As for comm.... Well that's not been very smooth either, and the ipykernel
dep is still carried by ipywidgets.

jtpio:
@martinRenou looks like the `8.1.0` release of package of widgets means we could start using the `comm` package here? https://github.com/jupyter-widgets/ipywidgets/releases/tag/8.1.0

martinRenou:
Yes!

bollwyvl:
Better do something, because https://github.com/jupyter-widgets/ipywidgets/pull/3811#issuecomment-1663249084

PR #34, opened 2023-03-22
Update to `jupyterlite-core==0.1.0b22`

jtpio:
Update to the latest release: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.1.0b20

- [x] Update to `jupyterlite-core==0.1.0b22`
- [x] Disabling `@jupyterlite/javascript-kernel-extension` is not needed anymore

PR #20, opened 2023-03-07
Update to Pyodide 0.23.0a1

jtpio:
Port of https://github.com/jupyterlite/jupyterlite/pull/962

image

PR #39, opened 2023-03-31
Update to Pyodide 0.23, update mocks

bollwyvl:
## References

- includes #20

## Changes

- [x] Update to `jupyterlite-core >=0.1.0b22`
- [x] bump pyodide all over (hopefully)
- [x] tests 
  - [x] fix fixtures (see https://github.com/jupyterlite/pyodide-kernel/pull/34#issuecomment-1480897879)
  - [x] restore some repo checks lost with doit (grabbed from #27) 
- [x] bump mocks
  - [x] `widgetsnbextension` shim (see https://github.com/jupyterlite/pyodide-kernel/issues/37#issuecomment-1490817197)
  - [x] update `requires-python` for all

## Follow-on

- investigate whether `micropip 0.3.0` features might simplify some things


bollwyvl:
Everything working like a boss on RTD (such as it is), just the one deprecation to fix. Widgets looking fine.

There's probably some more things to explore, but it's probably worthwhile to get this onto main (or even out as a pre) so we can start banging on it in more places. Will probably do a lite PR pulling in wheels from here, maybe resurrect the conda/wheel resolving stuff so we can get a clear look at the demos before calling it done-done.

bollwyvl:
Looking fine over on:

https://jupyterlite--1031.org.readthedocs.build/en/1031/_static/lab/

jtpio (approved):
Thanks!

PR #42, opened 2023-04-04
🎛 added `pipliteWheelUrl` option

stevejpurves:
This adds an option to set the `pipliteWheelUrl` from PageConfig in `litePluginSettings` allowing it to be set externally.

So intended usage would be something like this (or equivalent in the `juptyer-lite.json` file)

```
  PageConfig.setOption(
    'litePluginSettings',
    JSON.stringify({
      '@jupyterlite/pyodide-kernel-extension:kernel': {
        pipliteUrls: ['https://unpkg.com/@jupyterlite/pyodide-kernel@0.0.6/pypi/all.json'],
        pipliteWheelUrl: 'https://unpkg.com/@jupyterlite/pyodide-kernel@0.0.6/pypi/piplite-0.0.6-py3-none-any.whl'
      },
    }),
  );
```

I'm not clear on how to further test this option? either via the PR preview or locally via the cli.

### DIFF
--- a/packages/pyodide-kernel/src/kernel.ts
+++ b/packages/pyodide-kernel/src/kernel.ts
@@ -62,7 +62,7 @@ export class PyodideKernel extends BaseKernel implements IKernel {
       baseUrl,
       pyodideUrl,
       indexUrl,
-      pipliteWheelUrl: options?.pipliteWheelUrl || pipliteWheelUrl.default,
+      pipliteWheelUrl: options.pipliteWheelUrl || pipliteWheelUrl.default,
       pipliteUrls,
       disablePyPIFallback,
       location: this.location,
--- a/packages/pyodide-kernel/src/kernel.ts
+++ b/packages/pyodide-kernel/src/kernel.ts
@@ -62,7 +62,7 @@ export class PyodideKernel extends BaseKernel implements IKernel {
       baseUrl,
       pyodideUrl,
       indexUrl,
-      pipliteWheelUrl: options.pipliteWheelUrl ?? pipliteWheelUrl.default,
+      pipliteWheelUrl: options?.pipliteWheelUrl || pipliteWheelUrl.default,
       pipliteUrls,
       disablePyPIFallback,
       location: this.location,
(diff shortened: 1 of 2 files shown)


stevejpurves:
This will work for now for this use case and will be of immediate use in [this thebe PR](https://github.com/executablebooks/thebe/pull/612) but is a bit awkward, as the `all.json` already referred to in the `pipliteUrls`, already has the wheel specified. Would it be better to maybe work towards expose a way to override `allJSONUrl` and specify where to look up all the wheels on one (or more) "pypi-warehouse-like" `all.json` files?

jtpio:
Hmm I would have thought the existing `pipliteUrls` logic provided by this kernel would already be enough for fetching arbitrary wheels? Since they should be referenced via a URL as well.

stevejpurves:
@jtpio yes it does for arbitrary wheels. 

But in this case the change is for the *piplite wheel itself*, which does not look to `pipliteUrls` and instead looks to a fixed default path, as far as I understand it.

where a deployment does not want to concern itself with hosting any wheels at all (or more the issue having to reconfigure it's apps url-pathname routing to allow the wheels to be served at the default location) this change is needed to point to the piplite wheel itself. With that then all wheels needed can be hosted elsewhere or flexibly deployed.

packages/pyodide-kernel/src/kernel.ts
```diff
@@ -62,7 +62,7 @@ export class PyodideKernel extends BaseKernel implements IKernel {
       baseUrl,
       pyodideUrl,
       indexUrl,
-      pipliteWheelUrl: pipliteWheelUrl.default,
+      pipliteWheelUrl: options?.pipliteWheelUrl || pipliteWheelUrl.default,
```
jtpio, line 65: Wondering if the `?` is needed here, since `options` is not optional?

```suggestion
      pipliteWheelUrl: options.pipliteWheelUrl || pipliteWheelUrl.default,
```

jtpio (approved):
Thanks!

jtpio:
FYI @stevejpurves this should be available in `0.0.7`: https://github.com/jupyterlite/pyodide-kernel/releases/tag/v0.0.7

PR #45, opened 2023-04-20
Update to Pyodide `0.23.1`

jtpio:
Update to the latest Pyodide release: https://pyodide.org/en/stable/project/changelog.html#version-0-23-1


bollwyvl:
Seems fine on RTD: looks like there really wasn't much user-facing that changed, but still worth doing.

jtpio:
Yes, makes sense to stay on the latest stable :+1:

PR #47, opened 2023-05-05
fix: remove export kernel settings schema json

echarles:
This removes

```js
export * as KERNEL_SETTINGS_SCHEMA from '../schema/kernel.v0.schema.json';
```

which does not seem to be needed and fails downstream repos building with the error `...and cannot be used with 'export *'.`.

### DIFF
--- a/packages/pyodide-kernel-extension/src/index.ts
+++ b/packages/pyodide-kernel-extension/src/index.ts
@@ -12,7 +12,6 @@ import {
 import { IKernel, IKernelSpecs } from '@jupyterlite/kernel';
 import { IBroadcastChannelWrapper } from '@jupyterlite/contents';
 
-export * as KERNEL_SETTINGS_SCHEMA from '../schema/kernel.v0.schema.json';
 import KERNEL_ICON_SVG_STR from '../style/img/pyodide.svg';
 
 const KERNEL_ICON_URL = `data:image/svg+xml;base64,${btoa(KERNEL_ICON_SVG_STR)}`;


echarles:
@jtpio @bollwyvl https://jupyterlite-pyodide-kernel--47.org.readthedocs.build/en/47/_static/lab/index.html is working fine.

jtpio:
> fails downstream repos building with the error `...and cannot be used with 'export *'.`.

You mean in a downstream repo reusing the extension, but without webpack?

echarles:
it is failing while loading the node_modules while building. I am still trying to see why this is happening, and won't be able to link to a reproducer easily. Is that reexport needed?

911432:
It refers to this (site)[https://typescript.tv/errors/#ts2498].
But I don't fixed (this)[https://github.com/jupyterlite/pyodide-kernel/blob/main/packages/pyodide-kernel/src/_pypi.ts].
Because this is created automatically.

911432:
I would like to change it like this:
```typescript
export { default as KERNEL_SETTINGS_SCHEMA } from '../schema/kernel.v0.schema.json';
```

911432:
It refers to this [site](https://typescript.tv/errors/#ts2498).
I would like to change it like this:
```typescript
export { default as KERNEL_SETTINGS_SCHEMA } from '../schema/kernel.v0.schema.json';
```
But I don't fixed [this](https://github.com/jupyterlite/pyodide-kernel/blob/0.3.x/packages/pyodide-kernel/src/_pypi.ts).
Because this is created automatically.

911432 (changes requested):
It refers to this [site](https://typescript.tv/errors/#ts2498).
I would like to change it like this:
```typescript
export { default as KERNEL_SETTINGS_SCHEMA } from '../schema/kernel.v0.schema.json';
```
But I don't fixed [this](https://github.com/jupyterlite/pyodide-kernel/blob/0.3.x/packages/pyodide-kernel/src/_pypi.ts).
Because this is created automatically.

Issue #29, opened 2023-03-13
Add minimal novel package discovery


bollwyvl:
### Problem

At present, it is hard to calculate the exact package versions and URLs needed to build a reproducible site.

### Proposed Solution

Given a kernel session has installed a couple packages, offer something like `piplite.freeze` which would:
- call `micripip.freeze`
- weed out all of the packages that _weren't_ actually installed (not everyone needs `galpy` and `scipy`, but some _will_)
- inject the packages needed to get up to `pyodide-kernel`
- emit a `repodata.json` that can be dropped into `{lite_dir}/pypi/repodata.json`
- enjoy getting `import foo` without having to `%pip install foo`

### Additional context

- extracted from https://github.com/jupyterlite/pyodide-kernel/pull/27#issuecomment-1466231352

A much-heavier-duty version of this might try to use `micropip`, `depfinder` and/or `conda-lock` (which bundles poetry) to do this in a browserless, nodejs-less command line setting, but interactive-first is never the wrongest way to go.

rth:
Yes, I  confirm this functionality would be very useful to smooth the difference between notebooks with in-brower and out of browser kernels.

For instance, with the deployment of Jupyterlite in the scikit-learn example gallery (dev deployment only for now), it's a bit confusing to users why some packages can be just imported and others need to be pip installed first.

rth:
> inject the packages needed to get up to pyodide-kernel

Do you mean copy the packages from wherever they are into the deployed site? For reproducibility yes, that would be good. 

However, there is also a use case in the other direction where you want to add more PyPI packages to the `pyodide-lock.json` (pointing to the remote URLs where they are) without actually copying those files locally. For instance, scikit-learn deploys docs into a Github repo (that gets then deployed via Github Pages) and they don't realistically want to push 10+MB of packages there.

PR #49, opened 2023-06-22
Add Pyodide lockFileURL option

rth:
Related to the discussion https://github.com/pyodide/pyodide/issues/3573  by @bollwyvl once we have the ability to create a custom `pyodide-lock.json` we need to have some way to point this kernel to use it.

Supposing this `pyodide-lock.json`  is created alongside other JupyterLite files and deployed with those, I'm also not sure how that URL should look like: just a relative path?

### DIFF
--- a/packages/pyodide-kernel/src/tokens.ts
+++ b/packages/pyodide-kernel/src/tokens.ts
@@ -48,6 +48,11 @@ export namespace IPyodideWorkerKernel {
     indexUrl: string;
 
     /**
+     * The URL from which Pyodide will load the pyodide-lock.json file
+     */
+    pyodideLockFileURL: string;
+
+    /**
      * The URL of the `piplite` wheel for bootstrapping.
      */
     pipliteWheelUrl: string;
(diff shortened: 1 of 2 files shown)


rth:
Though looking at https://github.com/jupyterlite/pyodide-kernel/issues/29 @bollwyvl are you saying that `{lite_dir}/pypi/repodata.json` (now `{lite_dir}/pypi/pyodide-lock.json`) would be a hardcoded path, and that this configurability is not needed? Either way should be fine.

bollwyvl:
As of #103, `pyodideLockURL` (and JSON-compatible friends in `loadPyodide(options)`) are now available to be configured in [`jupyter-lite.json`](https://github.com/jupyterlite/pyodide-kernel/blob/main/examples/jupyter-lite.json#L6), which should ship as part of `jupyterlite-pyodide-kernel 0.3.0`.

Minor upstream note: moving the `options` typing into a named interface would make this just _that_ much better for downstreams! I didn't fully grok the ts-in-js toolchain to make that happen, however...

Looking further down the line, these options should put enough pieces in place to support an addon that: 
- uses e.g. a browser to do a real solve of the lock from `micropip.freeze` 
- grabs all the wheels
- backfills the import names with `pyodide-lock` 

I had started them in the same PR, but it became clear that this would add rather a lot of complexity, so will probably try to land it as another installable package, but theoretically could become an `[lock]` extra of this package.

PR #51, opened 2023-08-02
Add `before-bump-version` hook

jtpio:
Investigating the CI failure from https://github.com/jupyterlite/pyodide-kernel/pull/50

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -136,6 +136,10 @@ addopts = [
 version_cmd = "python scripts/bump-version.py"
 
 [tool.jupyter-releaser.hooks]
+before-bump-version = [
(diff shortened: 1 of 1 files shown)


jtpio:
The workflow was picking up the default `lerna` installed on the runners:

image

Lerna 7 is installed by default on the GitHub Actions runners, but we are still on version 6 at the moment:

https://github.com/jupyterlite/pyodide-kernel/blob/af3c4410cf22bd6c23a3bb3a2b9b0bde479f41d9/package.json#L54

https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#project-management

image

PR #53, opened 2023-08-03
Add `version_info` to the `ipykernel` mock

jtpio:
To help support the new ipywidgets 8.1.0.

Fixes https://github.com/jupyter-widgets/ipywidgets/issues/3819

image

PR #54, opened 2023-08-03
Make use of the new Comm package

martinRenou:
This goes in the direction of removing the ipykernel shim in the (how long?) long run.

What this PR does:
- installs the new `comm` package and make use of the new way to provide comms
- cleanup the ipykernel.comm shim, we should keep this one around for some time, for backward compat. But this ipykernel module **is not imported by default anymore**, meaning we don't end up in [this condition](https://github.com/jupyter-widgets/ipywidgets/pull/3811/files#diff-49cb3f90878a6146a8561bb1e55aa35c0667e11ab6446019780eb1f1b96a18a8R8) in ipywidgets anymore. If users import ipykernel before importing ipywidgets, [Jeremy's PR](https://github.com/jupyterlite/pyodide-kernel/pull/53) will make sure the ipywidgets code keeps running without failure and comms will be imported through our ipykernel shim.


martinRenou:
Working nicely 

Screenshot from 2023-08-03 10-36-17

jtpio (approved):
Nice thanks!!

Technically we could go with `0.0.11` for the next release if it's still backwards compatible. But maybe we could bump to `0.1.0` just to be sure?

martinRenou:
This should be backward compatible yes. But 0.1.0 just to be sure sounds good too.

PR #55, opened 2023-08-22
Bumped @jupyterlite packages

andeplane:
In https://github.com/jupyterlite/jupyterlite/pull/1075, we introduced a breaking change on the broadcasting system to prevent multiple tabs to communicate between each other in a way that crashed. But since the [pyodide-kernel](https://github.com/jupyterlite/pyodide-kernel/blob/main/packages/pyodide-kernel/src/worker.ts#L6) uses DriveFS from the [@jupyterlite/contents package](https://github.com/jupyterlite/pyodide-kernel/blob/main/packages/pyodide-kernel/package.json#L55) which currently points to `0.1.0-beta.18`, we are currently not in sync which can cause file Pyodide kernel to crash, see attached screenshot

This happens because pyodide asks the pyodide-kernel about some information about file system, and the worker sends a message to the jupyterlite main thread, but it is missing the `receiver: 'broadcast.ts'` part of the message. This also explains why I had issues with the other PR producing non working notebooks in the UI tests.

image

### DIFF
--- a/packages/pyodide-kernel-extension/package.json
+++ b/packages/pyodide-kernel-extension/package.json
@@ -48,10 +48,10 @@
   },
   "dependencies": {
     "@jupyterlab/coreutils": "^5.5.2",
-    "@jupyterlite/contents": "^0.1.0-beta.18",
-    "@jupyterlite/kernel": "^0.1.0-beta.18",
+    "@jupyterlite/contents": "^0.1.2",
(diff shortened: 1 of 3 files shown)


jtpio (approved):
Thanks!

jtpio:
Starting a new release now.

PR #58, opened 2023-09-13
Support jupyterlite-core 0.2.0a0

### LINKED ISSUE #57: Support jupyterlab-core >=0.2.0a0
### Problem

While things _work_ on the upstream RTD, the version pins likely won't let it work.

### Proposed Solution

- update to the new versions
- probably don't try to support `0.1.0` and `0.2.0` in the same code
- start a `0.1.x` branch

### Additional context

...

bollwyvl:
## References
- fixes #57

## Changes
- [x] update pins
  - > `prettier` is... undecided on trailing commas for this PR. none of the settings provide a stable output :crying_cat_face:  
- [x] collapse more root `./.dotwhatever` noise into `./build/.cache`
- [x] update all the versions
  - > i know the releaser, version bumpers, etc. love me and want me to be happy, but it's really hard to reason about this stuff vs new, nested, semi-undefined yarn/nx/lerna/esbuild/hatch/gha behavior
  - > either way, it shouldn't make a difference if we skip releasing `a0` 
  - among these, the `widgetsnbextension` stuff needs to be backported to `0.1.x`, as it is causing active breakage downstream vs ipywidgets `8.1.1`
    - #62 
    - #63
- [x] update docs
  - [x] add compatibility matrix and prerelease install instructions
- [ ] ??? 

## Downstreams

The hot sdist and wheel are up on the RTD preview:

```
https://jupyterlite-pyodide-kernel--58.org.readthedocs.build/en/58/_static/jupyterlite_pyodide_kernel-0.2.0a0.tar.gz
https://jupyterlite-pyodide-kernel--58.org.readthedocs.build/en/58/_static/jupyterlite_pyodide_kernel-0.2.0a0-py3-none-any.whl
```


jupyterlite_pyodide_kernel/_version.py
```diff
@@ -1,2 +1,2 @@
 """source of truth for ``jupyterlite-pyodide-kernel``` version."""
-__version__ = "0.1.1"
+__version__ = "0.2.0a0"
```
jtpio, line 2: Bumping the version is done when making the release with the releaser so we can revert that changes and other bumps in the `package.json`: https://github.com/jupyterlite/pyodide-kernel/blob/main/RELEASE.md#specifying-a-version-spec

bollwyvl: yep, i know: see above. I wanted to actually see something working with the interplay between the new versions.

and the `check-release` is showing it can handle going up to `0.2.0a1`.

jtpio: Not sure to understand what bumping the version in this PR helps with?

bollwyvl: It is for _my_ understanding, and presumably `lerna/yarn/esbuild/webpack`'s, as i was having issues with cache collisions, pulling the old packages, etc. even after blowing everything away, new envs all that.

also this leaves accurately-numbered wheels/sdist in the RTD build, as i'm trying to do downstream testing as well, due to the limited surface in this test suite and docs site.

and again, the bumper does fine on check-release, correctly going to `0.2.0a1`, so won't hurt anybody.

jtpio: > also this leaves accurately-numbered wheels/sdist in the RTD build

Are they not already "scoped" per sub-domain, for example https://jupyterlite-pyodide-kernel--58.org.readthedocs.build

README.md
```diff
@@ -16,7 +16,7 @@
 ## Requirements
 
 - `python >=3.8`
-- `jupyterlite >=0.1.0b19`
+- `jupyterlite-core >=0.2.0a0`
```
jtpio, line 19: Technically end users can still stick to JupyterLite 0.1.0 for now.

Or maybe the README should mention `jupyterlite-core >=0.2.0a0` for `jupyterlite-pyodide-kernel >=0.2.0a0`.

jupyterlite_pyodide_kernel/_version.py
jtpio: Should the version bump be reverted if downstream testing is done now?

README.md
bollwyvl: Yep, need to start the compatibility table (bleah).

bollwyvl: 09b9e75

PR #70, opened 2023-09-29
[0.1.x] bump core to 0.1.3

bollwyvl:
Not sure all of the things that landed, upstream: this is mostly to see if it fixed some of the known issues on RTD.


bollwyvl:
Yep, looking good on RTD... wouldn't strictly require a release just yet.

jtpio (approved):
Thanks!

PR #82, opened 2024-02-03
Update to pyodide 0.25.0

### LINKED ISSUE #81: support pyodide 0.25
### Problem

Support [pyodide 0.25.0](https://github.com/pyodide/pyodide/releases/tag/0.25.0).

### Proposed Solution

- known
  - [ ] bump CDN url
  - [ ] bump typings in `package.json`
- unknown
  - [ ] review changes to vendored `micropip`, etc.
  - [ ] really look hard at `pyodide-lock.json`, and whether we can migrate to it
    - [ ] _still_ doesn't support multiple versions

### Additional context

- ???

bollwyvl:
## References
- fixes #81 

## Changes
- [x] pyodide 0.25.0
  - [x] bump versions all over
- [x] lint
  - [x] run linters
    - > in a fresh env, not sure what we were running before
  - [x] replace `black` with `ruff format`
  - [x] also apply `ruff` to `examples/*.ipynb`
- [x] licenses
  - [x] tighten up all `(c) 2022` to `(c)` so we don't have to again
    - > seems like the right play, `jupyterlab` upstream has been doing as well

## Upstream notes

Looked through the [upstream changes](https://github.com/pyodide/pyodide/blob/0.25.0/docs/project/changelog.md#general) and didn't find anything that looked like a showstopper. 
- `await input()` does not appear to have changed
- I didn't observe a measurable difference in load time, as the hard loading of `openssl` and `ssl` (1.5mb) is probably dominating any micro-optimizations

## Not-on-this-pr

- [ ] proper support for custom/local pyodide lockfiles vs arbitrary wheels
  - started a little bit down this road, but don't think we should to do it on this PR, or potentially on this release.


jtpio (approved):
Nice, thanks for handling this update!

jtpio:
> Looked through the [upstream changes](https://github.com/pyodide/pyodide/blob/0.25.0/docs/project/changelog.md#general) and didn't find anything that looked like a showstopper.

Right, this can likely go in a `0.2.2` release then it seems.

bollwyvl:
Related to the _not-on-this-pr_ work: added https://github.com/jupyterlite/pyodide-kernel/issues/83

jtpio:
> Ugh, on `main`:

That looks related to the JupyterLab 4.1.0 release. And likely an issue in JupyterLab, as a minor version should not require extensions to update their dependencies.

bollwyvl:
Yeah `4.1.0` is indeed matched by `^4.0.7`, to my knowledge.

PR #84, opened 2024-02-05
pin to jupyterlab <4.1.0 for now in build-system and dev

bollwyvl:
## References
- continues #82

## Changes
- tighten pin of `jupyterlab` to `<4.1.0` in `build-system` and the `[dev]` extra
  - avoids [`Extensions require a devDependency on @jupyterlab/builder@^4.1.0, you have a dependency on 4.0.12`](https://github.com/jupyterlite/pyodide-kernel/actions/runs/7783301931/job/21221482902#step:9:99)
- specify `nodejs 20.x` and `python 3.11` in CI actions

### DIFF
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -17,6 +17,9 @@ jobs:
 
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
+        with:
+          node_version: '20.x'
+          python_version: '3.11'
 
       - name: Check Release
         uses: jupyter-server/jupyter_releaser/.github/actions/check-release@v2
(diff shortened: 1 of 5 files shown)


jtpio (approved):
Thanks!

jtpio:
Following-up with the `0.2.2` release.

bollwyvl:
https://github.com/conda-forge/jupyterlite-pyodide-kernel-feedstock/pull/16

jtpio:
As @krassowski pointed out on Gitter, the failure was probably related to the use of `~` here:

https://github.com/jupyterlite/pyodide-kernel/blob/996ac7ce1bfb9d1bb7f7b1386565c506eef82fa1/packages/pyodide-kernel-extension/package.json#L57

Which in a way makes sense, because of the overlap check JupyterLab performs (in https://github.com/jupyterlab/jupyterlab/blob/82051a9d0c6dc93677b0b66560aea48d71aaa043/jupyterlab/commands.py#L2252).

Also other extensions using `^` to define their dependencies are likely not affected.

Issue #91, opened 2024-02-15
Information to build custom wheels compatible with every pyodide-kernel versions


ManonMarchand:
Hi here! 

### Problem

I'm wondering if it would be possible to document the version of pyodide that corresponds to each version of pyodide-kernel ?

It's in constants.py but I could'nt find it somewhere else.

Here the change from pyodide-kernel 0.2.0 to 0.2.3 went from pyodide 0.24.1 to 0.25.0. This included breaking changes downstream. For example pyodide 0.24 is for python 3.10 and pyodide 0.25 is for python 3.11 so every custom wheels had to be rebuilt.

### Suggested Improvement

Maybe that's something that could be in the readme? Or in the shipping your own wheels part of the documentation here : https://jupyterlite.readthedocs.io/en/stable/howto/pyodide/wheels.html

I'd think something like a table giving all the information needed to build a wheel compatible with each version of pyodide-kernel

| pyodide-kernel version | pyodide version | python version | emscripten version
|--------|--------|--------|--------|
| 2.2.0 | 0.24.1 | 3.10 | 3.1.45
| 2.2.3 | 0.25 | 3.11 | 3.1.46 

What do you think?

jtpio:
Thanks @ManonMarchand for opening this issue :+1: 

Yes such table would be a good addition to the docs! Probably it's fine to keep it here in this repo (either in the README or in the docs), and then link to it from https://jupyterlite.readthedocs.io/en/stable/howto/pyodide/wheels.html ?

> Here the change from pyodide-kernel 0.2.0 to 0.2.3 went from pyodide 0.24.1 to 0.25.0. This included breaking changes downstream. For example pyodide 0.24 is for python 3.10 and pyodide 0.25 is for python 3.11 so every custom wheels had to be rebuilt.

For that particular case maybe we should have bumped `pyodide-kernel` to `0.3.0` just to be sure.

PR #92, opened 2024-02-15
docs: add information on compatibility with pyodide versions

ManonMarchand:
Following the discussion in #91, this adds a table of compatibility between `jupyterlite-pyodide-kernel` and `pyodide`. 

The last two columns could be ommited as one can access the information by looking through pyodide tags. I find it useful, but it sets the maintenance cost is a bit higher for each release of pyodide-kernel. Your choice.


ManonMarchand:
(force pushed to make prettier happy)

jtpio (approved):
Thanks!

Issue #40, opened 2023-03-31
Investigate new micropip features


bollwyvl:
### Problem

We still carry a lot of patches, managed as full-fat wheel builds, #27 adds _more_.

### Proposed Solution

Investigate whether new micropip features make this more reasonable.

### Additional context

extracted from #39

bollwyvl:
The `micropip` options were a bit too complex to solve in the otherwise-small #103. While these could still be added to `piplite` configuration somehow, it may be moot if we can get a closed form, build-time solution that generated a well-formed `pyodide-lock.json`, currently under investigation.

bollwyvl:
Ok, I've got an (unreleased) [jupyterlite-pyodide-lock](https://github.com/bollwyvl/jupyterlite-pyodide-lock) up to a working state. There are some temporary [docs](https://bollwyvl.github.io/jupyterlite-pyodide-lock/) with demo up, which also includes a deployed [lockfile](https://bollwyvl.github.io/jupyterlite-pyodide-lock/_static/static/pyodide-lock/pyodide-lock.json) and [wheel](https://bollwyvl.github.io/jupyterlite-pyodide-lock/_static/jupyterlite_pyodide_lock-0.1.0a0-py3-none-any.whl).

Right now it works with well-known locations of `firefox` and `chrom(e|ium)` over either: 
- a naive `subprocess` (sadly couldn't reuse much of the cpython gubbins, as they punt to `open`, etc)
- the [WebDriver](https://www.w3.org/TR/webdriver/) standard, via `selenium`
  - this is separate installable package in `contrib`

Due to the hard dependency on (alpha) `pyodide-lock`, and the general flakiness of browser automation, this should probably _not_ land in `pyodide-kernel` itself any time soon, though likely _should_ be mentioned in the appropriate docs.

Other techniques (`nodejs`, `playwright`, `pyppetteer`, an embedded `qt` browser) are certainly _possible_, and could be added via the `jupyterlite_pyodide_lock.locker.v0` entrypoint, but... I personally have no interest in maintaining them.

bollwyvl:
and of course not precious about where this actually gets released _from_, if it makes more sense for it to be in the org.

PR #103, opened 2024-03-26
Support `loadPyodide` options

bollwyvl:
## references
- for #40
- part of intent #49

## code changes
- [x] adds the schema for a subset of `loadPyodide(options)`
  - [x] some typing for the two initial interesting fields, `lockFileURL` and `packages`
- [x] only `piplite.install` packages that weren't preloaded by `loadPyodideOptions.packages`
- [x] update example to show use of this for e.g. `ssl`, `matplotlib`, etc.

## user-facing changes
- site owners will be able to further control the JSON-compatible parts of `loadPyodide`
- it's _possible_ preloading the `packages` makes things faster, as they are downloaded while the pyodide VM is starting up


bollwyvl:
Up on RTD, it does indeed look like it preloads `packages` from the pyodide distribution  e.g. `matplotlib`. 

This leads to an (apparently) slower start of the first interactive use of the kernel, so may not desirable for heavyweight packages on general purpose sites, but probably _will_ improve start time when we can skip `micropip` resolution for all of the `ipython` deps.

tomjakubowski:
Is this expected to make it in the upcoming 0.3.0 kernel release?

jtpio:
`0.3.0` is now available: https://github.com/jupyterlite/pyodide-kernel/releases/tag/v0.3.0

PR #105, opened 2024-03-30
Fix URL rewriting in `loadPyodideOptions`

bollwyvl:
## references
- continues #40

## code changes
- [x] fix rewriting of `*URL` values in `loadPyodideOptions`
  - > is currently trying to resolve the `key` instead of the `value`
- [x] add a static URL to demonstrate on RTD
  - > right now, just adds `?from-lite-config=1`, as we don't have a local lock handy 
- [x] fix typing
  - > `lockFileURL`, not `pyodideLockURL`

## user-facing changes
- site owners will be able to specify a `loadOptions.pyodideLockURL` (and _maybe_ other URLs) and have them resolve correctly

## notes
- will also be checking this against a relative URL to an in-tree `pyodide-lock.json`, might need some more info in the `URI` constructor.
  - oof. relative URLs in the lockfile are calculated against `indexURL`.
  - for a locally-managed `pyodide-lock.json` to work, it will need to be deployed along with (at least) a `pyodide.js`
    - however, it _won't_ have to ship the entire ~200mb distribution (if not wanted), as absolute URLs are respected

### DIFF
--- a/examples/jupyter-lite.json
+++ b/examples/jupyter-lite.json
@@ -5,7 +5,8 @@
     "litePluginSettings": {
       "@jupyterlite/pyodide-kernel-extension:kernel": {
         "loadPyodideOptions": {
-          "packages": ["matplotlib", "micropip", "numpy", "sqlite3", "ssl"]
+          "packages": ["matplotlib", "micropip", "numpy", "sqlite3", "ssl"],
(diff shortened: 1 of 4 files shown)


jtpio (approved):
Thanks!

bollwyvl:
Thanks. Making progress on the lock tool... But there's a lot of pointy
edges.

PR #106, opened 2024-05-21
Update to JupyterLite 0.4.0

jtpio:
Follow-up to https://github.com/jupyterlite/jupyterlite/releases/tag/v0.4.0a0

PR #107, opened 2024-05-21
Update `actions/setup-python`

jtpio:
Troubleshoot the macos CI failure, as seen in https://github.com/jupyterlite/pyodide-kernel/pull/106

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,8 +91,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.8'
-          architecture: 'x64'
+          python-version: '3.10'
 
       - uses: actions/download-artifact@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -89,10 +89,9 @@ jobs:
 
     steps:
       - name: Install Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
-          python-version: '3.8'
-          architecture: 'x64'
+          python-version: '3.10'
```
bollwyvl, line 94: hm, should this be `${{ matrix.python-version }}`?

jtpio: Right, I don't recall we had `3.8` specified before.

jtpio:
@meeseeksdev please backport to 0.3.x

PR #114, opened 2024-06-07
FileSystem calls over Atomics.wait instead of service worker when available

martinRenou:
This has the implication of moving from using `comlink` to using `coincident` for the Main <-> Worker communcication

### DIFF
--- a/packages/pyodide-kernel/src/worker.ts
+++ b/packages/pyodide-kernel/src/worker.ts
@@ -3,11 +3,45 @@
 
 import type Pyodide from 'pyodide';
 
-import type { DriveFS } from '@jupyterlite/contents';
+import { ContentsAPI, DriveFS, ServiceWorkerContentsAPI, TDriveMethod, TDriveRequest, TDriveResponse } from '@jupyterlite/contents';
 
 import { KernelMessage } from '@jupyterlab/services';
 
 import type { IPyodideWorkerKernel } from './tokens';
+
+
+/**
+ * An Emscripten-compatible synchronous Contents API using shared array buffers.
+ */
+export class SharedBufferContentsAPI extends ContentsAPI {
+  request<T extends TDriveMethod>(data: TDriveRequest<T>): TDriveResponse<T> {
+    return workerAPI.processDriveRequest(data);
+  }
+}
+
+/**
+ * A custom drive implementation which uses shared array buffers if available, service worker otherwise
+ */
+class PyodideDriveFS extends DriveFS {
+  createAPI(options: DriveFS.IOptions): ContentsAPI {
+    if (crossOriginIsolated) {
+      return new SharedBufferContentsAPI(
+        options.driveName,
+        options.mountpoint,
+        options.FS,
+        options.ERRNO_CODES,
+      );
+    } else {
+      return new ServiceWorkerContentsAPI(
+        options.baseUrl,
+        options.driveName,
+        options.mountpoint,
+        options.FS,
+        options.ERRNO_CODES,
+      );
+    }
+  }
+}
 
 export class PyodideRemoteKernel {
   constructor() {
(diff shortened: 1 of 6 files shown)


martinRenou:
Oh there are no visual regression tests testing the code execution. That makes me uncomfortable :)

martinRenou:
Well at least we have the readthedocs build for trying

packages/pyodide-kernel/src/worker.ts
```diff
@@ -1,509 +1,579 @@
...
 
-    if (options.location.includes(':')) {
-      const parts = options.location.split(':');
-      this._driveName = parts[0];
-      this._localPath = parts[1];
-    } else {
-      this._driveName = '';
-      this._localPath = options.location;
-    }
+  var kernel: any;
```
martinRenou, line 25: For the record, this is the reason of the `no-var` eslint rule removal. I'm not sure how to implement it differently.

jtpio: An alternative could be to use inline comments to disable the rules, per line or for the whole file?

bollwyvl: oh... this removes all of the class structure. This is not pleasing to me, as it would be all but impossible to overload by extension and add Just One Change to a relatively small method, rather than copy-pasta forking.

martinRenou: > further, the any make me really sad. Is it entirely unknowable?

Those variables were already `any`, it's not a change from this PR.

> oh... this removes all of the class structure. This is not pleasing to me, as it would be all but impossible to overload by extension and add Just One Change to a relatively small method, rather than copy-pasta forking.

I know... I'm not sure of a way to expose a class instance using `coincident`, I will look further

martinRenou:
Let's get https://github.com/jupyterlite/pyodide-kernel/pull/115 in first

packages/pyodide-kernel/src/worker.ts
martinRenou: Doing only the coincident transition and trying to keep the class def in https://github.com/jupyterlite/pyodide-kernel/pull/115

martinRenou:
I can't seem to be able to reproduce what you're seeing on this RTD build

Screenshot from 2024-06-17 15-00-17

jtpio:
> I can't seem to be able to reproduce what you're seeing on this RTD build

https://github.com/jupyterlite/jupyterlite/issues/1412 mentions the build on the JupyterLite repo, which likely uses an old release of `jupyterlite-pyodide-kernel`. In that case using https://jupyterlite--1383.org.readthedocs.build/en/1383/_static/lab/index.html I can also reproduce the error:

image

Could it be because that RTD demo uses the latest JupyterLite, but an older `jupyterlite-pyodide-kernel`. If so, wouldn't that be fixed with the JupyterLite 0.4.0 and `jupyterlite-pyodide-kernel` 0.4.0 final releases? (as with this PR)

bollwyvl:
Welp, i'm getting 100% repro of the above issue on linux/chromium 126.0.6478.61, in and out of an incognito window.

image

We _know_ it won't work in FF private, but also seeing in firefox 127.0:

image

martinRenou:
I've looked deeper into it, I believe https://github.com/jupyterlite/jupyterlite/pull/1418 should fix it. Can you give it a try? I see it working there

jtpio (approved):
Thanks!

jtpio:
FYI @bollwyvl `0.4.0b0` is out with this change: https://github.com/jupyterlite/pyodide-kernel/releases/tag/v0.4.0b0

If that looks good to you, we can likely move forward with JupyterLite 0.4.0 and release final in the coming days.

bollwyvl:
Won't be able to review much this weekend, but will roll it on conda-forge and see what we get...

PR #115, opened 2024-06-07
Switch from using comlink to coincident

martinRenou:
This will allow us using shared array buffers for filesystem access for free


jtpio:
For reference and to provide more context on why use `coincident`: [`synclink`](https://github.com/pyodide/synclink) would have been an option too, but it is not actively maintained, and `coincident` was recommended in https://github.com/pyodide/synclink/pull/20#issuecomment-1837217978

martinRenou:
Adding a note that there is a comlink dependency in Jupyterlite core that seems to be pulled only for a type definition, we may want to get rid of it

packages/pyodide-kernel/src/coincident.worker.ts
```diff
@@ -0,0 +1,25 @@
...
+import coincident from 'coincident';
+
+import { PyodideRemoteKernel } from './worker';
+import { IPyodideWorkerKernel } from './tokens';
+
+const worker = new PyodideRemoteKernel();
+
+const workerAPI: IPyodideWorkerKernel = coincident(self) as IPyodideWorkerKernel;
+
+workerAPI.initialize = worker.initialize.bind(worker);
```
bollwyvl, line 16: At the very least, we should sort these... but better still would be to: 
- enforce it with a type
- do this dynamically
- automate it with magic comments (ick)

packages/pyodide-kernel/src/kernel.ts
```diff
@@ -24,7 +24,7 @@ export class PyodideKernel extends BaseKernel implements IKernel {
     super(options);
     this._worker = this.initWorker(options);
     this._worker.onmessage = (e) => this._processWorkerMessage(e.data);
-    this._remoteKernel = wrap(this._worker);
+    this._remoteKernel = coincident(this._worker) as IPyodideWorkerKernel;
```
bollwyvl, line 27: `coincident` takes a second options argument, haven't dug into what they do. 

Perhaps this could be moved into an `protected _initRemote(options): IPyodideWorkerKernel` which can be overloaded by a downstream.

packages/pyodide-kernel/src/worker.ts
```diff
@@ -349,7 +349,7 @@ export class PyodideRemoteKernel {
    *
    * @param content The incoming message with the comm target name.
    */
-  async commInfo(content: any, parent: any) {
+  async commInfo(content: any, parent: any): Promise<{ comms: any; status: 'ok' }> {
```
bollwyvl, line 352: Would hoist to new named type, i guess.

packages/pyodide-kernel/src/coincident.worker.ts
jtpio: @martinRenou is this something that should be done as part of this PR, or could be done as a follow-up?

So we can move forward with this PR and https://github.com/jupyterlite/pyodide-kernel/pull/114

martinRenou: We can probably do this as a follow up?

PR #120, opened 2024-06-18
Update Dependency

911432:
You must do 'npm run quickstart' twice at first.
[pytest.zip](https://github.com/user-attachments/files/15886309/pytest.zip)
Then it works.
run


911432:
I'll do it later.

PR #126, opened 2024-07-12
Use `coincident` if `crossOriginIsolated`, `comlink` otherwise

jtpio:
Looking into https://github.com/jupyterlite/jupyterlite/issues/1424.

`coincident` uses a SharedArrayBuffer here: https://github.com/WebReflection/coincident/blob/01cc8a1e90abb360deb9eabd7773ef4355fcbc3d/esm/index.js#L82

However in environments that are not `crossOriginIsolated`, usings `coincident` on the worker side leads to the following issues:

```
TypeError: [object Int32Array] is not a shared typed array.
```

So this PR takes the following approach:

- use `coincident` on both the main thread and the worker if `crossOriginIsolated, to fix https://github.com/jupyterlite/jupyterlite/issues/1424 / #127 
- use `comlink` otherwise, this time on both the main thread and the worker too

Fixes https://github.com/jupyterlite/pyodide-kernel/issues/127

### DIFF
--- a/packages/pyodide-kernel/src/comlink.worker.ts
+++ b/packages/pyodide-kernel/src/comlink.worker.ts
@@ -9,11 +9,12 @@ import { expose } from 'comlink';
 
 import { ContentsAPI, DriveFS, ServiceWorkerContentsAPI } from '@jupyterlite/contents';
 
-import { PyodideRemoteKernel } from './worker';
 import { IPyodideWorkerKernel } from './tokens';
 
+import { PyodideRemoteKernel } from './worker';
+
 /**
- * A custom drive implementation which uses shared array buffers if available, service worker otherwise
+ * A custom drive implementation which uses the service worker
  */
 class PyodideDriveFS extends DriveFS {
   createAPI(options: DriveFS.IOptions): ContentsAPI {
--- a/packages/pyodide-kernel/src/comlink.worker.ts
+++ b/packages/pyodide-kernel/src/comlink.worker.ts
@@ -1,0 +1,60 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+/**
+ * A WebWorker entrypoint that uses comlink to handle postMessage details
+ */
+
+import { expose } from 'comlink';
+
+import { ContentsAPI, DriveFS, ServiceWorkerContentsAPI } from '@jupyterlite/contents';
+
+import { PyodideRemoteKernel } from './worker';
+import { IPyodideWorkerKernel } from './tokens';
+
+/**
+ * A custom drive implementation which uses shared array buffers if available, service worker otherwise
+ */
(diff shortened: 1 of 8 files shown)


jtpio:
Current status, after a naive replacement of `postMessage` by `coincident`:

image

bollwyvl:
Maybe we need to be using [`coincident/structured`](https://github.com/WebReflection/coincident?tab=readme-ov-file#coincidentstructured)?

jtpio:
Right, I'll have a closer look.

jtpio:
This seems to be caused by `coincident` using a `SharedArrayBuffer` here: https://github.com/WebReflection/coincident/blob/01cc8a1e90abb360deb9eabd7773ef4355fcbc3d/esm/index.js#L82

It seems to be working fine when serving the site with right headers, for example with: `npx static-handler --cors --coop --coep --corp ./build/docs-app`

I thought `coincident` would default to the `comlink` behavior when `SharedArrayBuffer` are not available, but it looks like it's not the case.

jtpio:
Switching to `coincident` on the worker side does seem to be fixing https://github.com/jupyterlite/jupyterlite/issues/1424.

[coincident-widget-error.webm](https://github.com/user-attachments/assets/00cacaa2-50cb-4ce1-82b8-bd62d40972a0) 

> I thought `coincident` would default to the `comlink` behavior when `SharedArrayBuffer` are not available, but it looks like it's not the case.

Maybe we need to clearly separate the case when `SharedArrayBuffer` are available on the page from when they are not. Meaning using `coincident` when they are, and `comlink` when they are not (on both the worker and the main threads).

cc @martinRenou in case you have some input on this

martinRenou:
> Meaning using coincident when they are, and comlink when they are not (on both the worker and the main threads).

That's unfortunate, but if it's the only solution than let's go for it.

coincident is really taking the burden of implementing a proper communication protocol using SharedArrayBuffers, so it's really useful to us in that case. That's unfortunate it cannot work in an asynchronous fashion though, closer to the comlink approach.

martinRenou:
Thanks for looking into it by the way 👍🏽 I'm swamped with other projects. We should apply similar changes to jupyterlite/xeus once we have a proper solution.

bollwyvl:
Yeah... As realizing the merit of the SAB is getting more and more tenuous,
it looks like it needs to be entirely opt-in.

jtpio:
I'll push a first draft towards that direction and will clean things up a bit afterwards.

packages/pyodide-kernel/src/comlink.worker.ts
jtpio, line 1: We could technically keep both entrypoints in a single file. But having a separate file for `coincident` and `comlink` may be easier to reason about.

martinRenou (approved):
Looks great!

jtpio:
Thanks for the review!

PR #129, opened 2024-07-22
Fix `mountDrive` check

jtpio:
Fixes https://github.com/jupyterlite/jupyterlite/issues/1429


jtpio:
Merging as @juntyr confirmed the fix in https://github.com/jupyterlite/jupyterlite/issues/1429#issuecomment-2243630465.

Issue #128, opened 2024-07-18
Test file system access over `SharedArrayBuffer`


jtpio:
As a follow-up to https://github.com/jupyterlite/pyodide-kernel/pull/126, we should test accessing files from the kernel works with both the `SharedBufferContentsAPI` and `ServiceWorkerContentsAPI`.

This could mean setting up basic UI tests, starting different servers:

- with the COOP / COEP headers
- without the headers

PR #130, opened 2024-07-22
Add UI tests

jtpio:
Similar to the `jupyterlite-xeus` repo: https://github.com/jupyterlite/xeus

This would help with https://github.com/jupyterlite/pyodide-kernel/pull/126.

Fixes https://github.com/jupyterlite/pyodide-kernel/issues/128


jtpio:
For now these tests are simple and don't include snapshots. They should at least cover the cases for when the COOP / COEP headers are set, and checking that executing cells and the files can be seem from the kernel.

We can always iterated and add more in separate PRs.

Issue #143, opened 2024-11-23
Bug/Improvement: Correct handling of message sending


martenrichter:
Recently a lot of bug reports regarding jupyter lite widget interactivity being broken:
https://github.com/jupyterlite/pyodide-kernel/issues/133
https://github.com/jupyterlite/jupyterlite/issues/1465
https://github.com/jupyter-book/thebe/issues/771
https://github.com/jupyter/try-jupyter/issues/54

I have ran jupyter lite for two days in the debugger (see here my post about it here:
https://github.com/jupyterlite/pyodide-kernel/issues/133
).

And I am now relatively confident that the incomplete implementation of ``get_parent` in `kernel.py` and the js side:
https://github.com/jupyterlite/pyodide-kernel/blob/b70f099252670535db4745b5abab680300dc84eb/packages/pyodide-kernel/py/pyodide-kernel/pyodide_kernel/kernel.py#L29

is causing the trouble. As the control and shell channel parent is mixing through this simplification. 
In contrast to:
https://github.com/ipython/ipykernel/blob/188f39c509f1056c3784aad6954498f31de955a3/ipykernel/kernelbase.py#L622

So I think that a patch for this behavior is required. And this may fix the issues. Do you agree?

Are you open for PR? (But I must say, I am very experienced in C++ and js (and a bit in ts), but not in python, but I think the changes do not require much understanding of the language details, so it is possible that I will give up, if I fail). Also I am not so sure how I can test the changes on pyodide-kernel alone. So input is appreciated.

P.S: I hope it is ok to open a separate issue, I thought in this way it is a bit more organized.

martenrichter:
Ok, I tried to implement it. But it was not the cause. Anyway, the wrong ordering of message with their kernel header is the cause. I also tried to reuse the parent headers from the initial execution attempt, but this did not work either  (xeus has always the same parent msg_id, so I gave it a try). I will continue....

martenrichter:
Ok, next thing, if I go after reverting back to the commit before it still works.
If I clean all temp directories (e.g., node_modules) and build and install again, it does not work. 
If I change back to the reverted head in git build again, it works.
This is weird and may point to a specific combination of package versions???

martenrichter:
Additional things:
1.) Use test code:
```
%pip install -q ipywidgets
from ipywidgets import interact

# Function to handle text input
def display_text(value=''):
    return f"Updated value: {value}"

# Create an interactive Text widget
[...]

PR #144, opened 2024-11-26
Fix: Ensure `executeRequest` waits for standard output to complete when using `comlink`

higoo-higoo:
# Description of Issue
When using `comlink`, `executeRequest` function sometimes finishes before the standard output is fully completed.
This behavior appears to be caused by the lack of synchronization between callback handling and the simple function logic in `comlink`.

issue

# How to Reproduce the Issue
1. `npm run quickstart`, `jlpm build` , `jlpm docs:lite`  with `"exposeAppInBrowser": true` and `jlpm serve`
2. Access [JupyterLite](http://127.0.0.1:8000/docs-app/lab/index.html).
3. Execute the following code in your browser's console:
```javascript
window.jupyterapp.serviceManager.kernels._kernelConnections.values().next().value.iopubMessage.connect((e, t) => {
    const c = t.content;
    
    if (c?.execution_state === 'busy') {
        console.log('+++++++execution start+++++++');
    }

    console.log('!!', t.content);
    
    if (c?.execution_state === 'idle') {
        console.log('-------execution finish-------');
    }
});
```
4. Execute any code in JupyterLite.

- Expected behavior: The log -------execution finish------- should only appear after all standard output is processed.
- Actual behavior: The log sometimes appears prematurely, before the standard output is fully handled.

# Changes Made
I identified the issue as being caused by an inconsistency in the execution order between callback handling and the simple function logic when using `comlink`.

To resolve this:

I implemented a mechanism where the worker sends a message through a callback at the end of the process .
The kernel now waits for this message before concluding the `executeRequest` function.
This ensures proper synchronization and prevents premature termination of `executeRequest`.

### DIFF
--- a/packages/pyodide-kernel/src/kernel.ts
+++ b/packages/pyodide-kernel/src/kernel.ts
@@ -198,6 +198,9 @@ export class PyodideKernel extends BaseKernel implements IKernel {
         );
         break;
       }
+      case 'execute_return': {
+        this._executed.resolve();
+      }
     }
   }
 
@@ -244,6 +247,8 @@ export class PyodideKernel extends BaseKernel implements IKernel {
     await this.ready;
     const result = await this._remoteKernel.execute(content, this.parent);
     result.execution_count = this.executionCount;
+    await this._executed.promise;
+    this._executed = new PromiseDelegate<void>();
     return result;
   }
 
@@ -340,6 +345,7 @@ export class PyodideKernel extends BaseKernel implements IKernel {
     | IRemotePyodideWorkerKernel
     | Remote<IRemotePyodideWorkerKernel>;
   private _ready = new PromiseDelegate<void>();
+  private _executed = new PromiseDelegate<void>();
 }
 
 /**
--- a/packages/pyodide-kernel/src/worker.ts
+++ b/packages/pyodide-kernel/src/worker.ts
@@ -311,6 +311,11 @@ export class PyodideRemoteKernel {
       publishExecutionError(results['ename'], results['evalue'], results['traceback']);
     }
 
+    this._sendWorkerMessage({
+      parentHeader: this.formatResult(this._kernel._parent_header)['header'],
+      type: 'execute_return',
+    });
+
     return results;
   }
 
--- a/packages/pyodide-kernel/src/worker.ts
+++ b/packages/pyodide-kernel/src/worker.ts
@@ -312,7 +312,6 @@ export class PyodideRemoteKernel {
     }
 
     this._sendWorkerMessage({
-      parentHeader: this.formatResult(this._kernel._parent_header)['header'],
       type: 'execute_return',
     });
 


packages/pyodide-kernel/src/kernel.ts
```diff
@@ -198,6 +198,9 @@ export class PyodideKernel extends BaseKernel implements IKernel {
         );
         break;
       }
+      case 'execute_return': {
```
jtpio, line 201: Wondering if such message should be named something else, or prefixed with an underscore or similar? Otherwise it looks like it may be an official message part of the Jupyter protocol, while it isn't.

bollwyvl: sure, `jpk:` would be solid. It's not really _private_, and a downstream may well want to react to it. We'd pay a minimal tax to do the `.split` but adding more fields at this time seems even worse and more breaking.

Most of these should be hanging off a `const` or `namespace` somewhere (but not `enum`: those can break in round-tripped JSON settings).

martenrichter:
Do you think this also solves the ipywidget problem, which is also related to the commit introducing coimlink:
See here, the last comment for my analysis:
https://github.com/jupyterlite/pyodide-kernel/issues/143
Or do you think it is a separate issue in comlink communication?

jtpio:
> Do you think this also solves the ipywidget problem, which is also related to the commit introducing coimlink:

Not sure yet, but probably worth trying with this link (preview for the PR): https://jupyterlite-pyodide-kernel--144.org.readthedocs.build/en/144/_static/lab/index.html

martenrichter:
Nope, it still persists. So it must be another race condition introduced by comlink (or something else in the comlink commit).

packages/pyodide-kernel/src/kernel.ts
higoo-higoo: So, should we rename the message to `jpk:execute_return` after all??

packages/pyodide-kernel/src/kernel.ts
```diff
@@ -244,6 +247,8 @@ export class PyodideKernel extends BaseKernel implements IKernel {
     await this.ready;
     const result = await this._remoteKernel.execute(content, this.parent);
     result.execution_count = this.executionCount;
+    await this._executed.promise;
+    this._executed = new PromiseDelegate<void>();
```
martenrichter, line 251: I have a stupid question. There is always one PromiseDelegate. The execute request starts before interacting with the promise_delegate. If two execution requests are started almost at the same time, what identifies which promise to resolve once the message arrives?

martenrichter:
Another question: Does it make sense to add this message mechanism? I think the diagnosis is that Comlink sometimes does not accurately handle calls to async functions in the worker. But this means that potentially, all async functions are affected by the same issue. So, it may be better to find the root cause of why Comlink is failing. It is not that simple and uses UUIDS for every message, but I am also not sure if it can handle async.

martenrichter:
The real problem is that comlink changes the ordering of the messages from "this._sendWorkerMessage" and the return of the worker's async methods. So, the _sendWorkerMessages arrive after the return from the async method that sets the parent header with the msg_id.

martenrichter:
Ok, I have found the underlying problem.
This line:
https://github.com/jupyterlite/pyodide-kernel/blob/fb9c10dc5d149849d04dffbc073680c64c988993/packages/pyodide-kernel/src/kernel.ts#L89
ends up in comlink:
https://github.com/GoogleChromeLabs/comlink/blob/fd4b52666b1ec62784f8b45cb1108c7e40bc481d/src/comlink.ts#L213
creating a separate MessageChannel for the callback used in _sendWorkerMessage.

This explains all the troubles, as the call to the worker is done via the normal worker's message ports, but the _sendWorkerMessage uses a separate Message Channel. The UAs do not guarantee that two separate channels are executed in the same order, but ordering is crucial for the Jupyter protocol.
I see the following options:
1.) Proceed forward in the spirit of this PR, and add to every call to the worker a completion message, but have for every call a separate Promise may be using a Map using some kind of unique id, maybe the message id.
2.) Find a way to use comlink so that _sendWorkerMessage does not use a separate channel from the other mechanism.
3.) Go back to using postMessage for comlink, maybe with an added id `jupyterMessage` so that comlink knows this is not a message it should process.

martenrichter:
> Many thanks @martenrichter for the thorough investigation!
> 
> > 2.) Find a way to use comlink so that _sendWorkerMessage does not use a separate channel from the other mechanism.
> 
> Right, if there is a way to do that with comlink and stick to the comlink API if possible, it would be great. Otherwise switching back to raw `postMessage` would likely be fine since it's more important to fix the actual user issue 👍

I have posted a PR for 3.). My follow-up investigation showed that using native comlink is only possible if I submit an upstream patch, which will be complex. I do not know how welcome this will be from the comlink team side.

Issue #133, opened 2024-08-06
ipywidgets does not work in some situations


MRYingLEE:
For the following simple ipywidgets code:

```Python
%pip install ipywidgets
# Import IPython widgets for interactive input
from IPython.display import display
import ipywidgets as widgets

# Create a text box for user input
prompt_box = widgets.Textarea(
    value='Enter your prompt here...',
    placeholder='Type something...',
    description='Prompt:',
    disabled=False
)

# Create a button to generate text
button = widgets.Button(description="Generate Text")

# Function to handle button click
def on_button_click(b):
    prompt = prompt_box.value
    generated_text = "generate_text(prompt)"
    print("Prompt:", prompt)
    print("Generated Text:", generated_text)

# Attach the function to the button click event
button.on_click(on_button_click)

# Display the text box and button
display(prompt_box, button)
```

I works well in other Jupyter environment, such as Colab (https://colab.research.google.com/).
![image](https://github.com/user-attachments/assets/f71c56f5-5580-4d41-a9d1-7adb5c5f3918)

But the button clicking does not show anything in Jupyterlite, such as here (https://jupyterlite-pyodide-kernel.readthedocs.io/en/latest/_static/lab/index.html).

Please check.
Thanks,

MRYingLEE:
I used kernelspy(https://github.com/jupyterlab-contrib/jupyterlab-kernelspy) to check the messages after the button was clicked, and I saw:
![image](https://github.com/user-attachments/assets/33427299-5f6a-494b-beaa-50c7ed9ccc41)

So it seems the messaging system worked well but for some reason the messages were not processed well.

MartinaMJ:
Hello. I have been facing this same problem. I noticed the kernel responds to the first few interactions but then stops responding. Did you find a solution or a possible cause?

pierre-haessig:
My jupyterlite deployment (e.g. https://pierre-haessig.github.io/pierre-notebooks/notebooks/?path=Frequency%20regulation%20nocode.ipynb) is also facing broken widget interaction, despite being up to date with jupyterlite-core==0.4.1 and jupyterlite-pyodide-kernel==0.4.2 just like the [jupyterlite demo](https://github.com/jupyterlite/demo/) as per PR https://github.com/jupyterlite/demo/pull/148

[...]

PR #148, opened 2024-12-15
Use postMessage instead of a comlink produced MessageChannel

martenrichter:
Closes #144 
Fixes https://github.com/jupyterlite/pyodide-kernel/issues/133
Fixes https://github.com/jupyterlite/jupyterlite/issues/1465
Fixes https://github.com/jupyter-book/thebe/issues/771
Fixes https://github.com/jupyter/try-jupyter/issues/54

### DIFF
--- a/packages/pyodide-kernel/src/comlink.worker.ts
+++ b/packages/pyodide-kernel/src/comlink.worker.ts
@@ -33,7 +33,7 @@ export class PyodideComlinkKernel extends PyodideRemoteKernel {
     super();
     this._sendWorkerMessage = (msg: any) => {
       // use postMessage, but in a format, that comlink would not process.
-      postMessage({ jMsg: msg });
+      postMessage({ _kernelMessage: msg });
     };
   }
 
--- a/packages/pyodide-kernel/src/comlink.worker.ts
+++ b/packages/pyodide-kernel/src/comlink.worker.ts
@@ -29,6 +29,14 @@ class PyodideDriveFS extends DriveFS {
 }
 
 export class PyodideComlinkKernel extends PyodideRemoteKernel {
+  constructor() {
+    super();
+    this._sendWorkerMessage = (msg: any) => {
+      // use postMessage, but in a format, that comlink would not process.
+      postMessage({ jMsg: msg });
+    };
+  }
+
   /**
    * Setup custom Emscripten FileSystem
    */
--- a/yarn.lock
+++ b/yarn.lock
@@ -3061,7 +3061,7 @@ __metadata:
     "@jupyterlite/kernel": ^0.4.5
     "@types/jest": ^29.5.4
     coincident: ^1.2.3
-    comlink: ^4.4.1
+    comlink: ^4.4.2
     esbuild: ^0.19.2
     jest: ^29.7.0
     pyodide: 0.26.4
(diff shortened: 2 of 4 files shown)


martenrichter:
This is an alternative to:
https://github.com/jupyterlite/pyodide-kernel/pull/144
and should fix the current problems, if comlink is used.
Should fix https://github.com/jupyterlite/pyodide-kernel/issues/143 and related reports.

packages/pyodide-kernel/src/comlink.worker.ts
```diff
@@ -29,6 +29,14 @@ class PyodideDriveFS extends DriveFS {
 }
 
 export class PyodideComlinkKernel extends PyodideRemoteKernel {
+  constructor() {
+    super();
+    this._sendWorkerMessage = (msg: any) => {
+      // use postMessage, but in a format, that comlink would not process.
+      postMessage({ jMsg: msg });
```
jtpio, line 36: Wondering if something a bit more explicit like `_kernelMessage` could also work? With or without the leading `_`, but having it could help communicate it's used for "internal" message handling?

yarn.lock
```diff
@@ -5941,13 +5941,20 @@ __metadata:
...
 
-"comlink@npm:^4.3.1, comlink@npm:^4.4.1":
+"comlink@npm:^4.3.1":
   version: 4.4.1
   resolution: "comlink@npm:4.4.1"
   checksum: 16d58a8f590087fc45432e31d6c138308dfd4b75b89aec0b7f7bb97ad33d810381bd2b1e608a1fb2cf05979af9cbfcdcaf1715996d5fcf77aeb013b6da3260af
   languageName: node
   linkType: hard
 
+"comlink@npm:^4.4.2":
```
jtpio, line 5951: Looks like the two versions of `comlink` could be deduplicated with:

https://github.com/jupyterlite/pyodide-kernel/blob/fb9c10dc5d149849d04dffbc073680c64c988993/package.json#L20

packages/pyodide-kernel/src/comlink.worker.ts
martenrichter: Ok, but then also the coincident worker needs to rename the change.

yarn.lock
martenrichter: Ok, I will try this. I am not so fluent with yarn, I normally use npm. I'll be back in the evening at my dev computer.

packages/pyodide-kernel/src/comlink.worker.ts
jtpio: I meant for the `jMsg` key, we could still keep `_sendWorkerMessage` as the method name:

```suggestion
      postMessage({ _kernelMessage: msg });
```

yarn.lock
jtpio: Running `jlpm deduplicate` at the top-level should normally perform the deduplication step.

packages/pyodide-kernel/src/comlink.worker.ts
martenrichter: Ok, got it. and it is also unique enough that comlink ignores it. (comlink ignores it if no id field is set...)

martenrichter:
OK, I have included the suggested changes. The deduplication affected more than just the comlink package, I hope this is fine.

jtpio:
Released: https://github.com/jupyterlite/pyodide-kernel/releases/tag/v0.4.6

PR #149, opened 2024-12-16
Update to JupyterLite 0.5.0

jtpio:
Update to https://github.com/jupyterlite/jupyterlite/releases/tag/v0.5.0a0

- [x] Bump Python dependency
- [x] Update `@jupyterlite` packages
- [x] Update `@jupyterlab/builder` package
- [x] Update README
- [x] Fix `@playwright/test`


bollwyvl:
Hm, with 0.4.5 out (:tada:) we follow your approach on core and split that off to an `0.4.x` in case something comes up. I'll take a look in a bit.

bollwyvl (approved):
Looking good! Ran through some stuff on RTD (like `virtual_drive.ipynb` from upstream) and everything looks pretty good. Getting this out and under abusive downstream testing may uncover more issues, but see no blockers thus far.

jtpio:
I see we have a `0.4.x` branch: https://github.com/jupyterlite/pyodide-kernel/tree/0.4.x

> Oh, one little thing: guess we should go ahead and do the general bump to `requires-python = ">=3.9"`, both here and upstream, but that's probably a separate issue.

Yes we can do that in this cycle :+1: (started https://github.com/jupyterlite/jupyterlite/pull/1537 in the main repo)

PR #151, opened 2024-12-17
Drop Python 3.8

jtpio:
Since Python 3.8 is EOL.

Update the test matrix to 3.9 - 3.13.


bollwyvl (approved):
Looks good (in whichever order vs #150)

PR #153, opened 2025-01-01
Update to Pyodide 0.27, and fix compatibility with `micropip` 0.8

agriyakhetarpal:
## Description

This pull request adds changes in order to use the [newly released Pyodide version `0.27.0`](https://github.com/pyodide/pyodide/releases/tag/0.27.0), and fixes `piplite` to be able to install wheels. Here is a summary of the changes performed:
- the position of the `index_urls` and `fetch_kwargs` arguments in the `_MP_QUERY_PACKAGE` `micropip` wrapper has been fixed, since `micropip` reversed them in [`micropip.package_index.query_package`](https://github.com/pyodide/micropip/blob/4301606cd4d98d13b766630cc36cf19a4b5fb643/micropip/package_index.py#L264) in versions 0.7 and later.
- `pyodide_YYYY_build_number_wasm32` has been added as a supported platform tag. Recent versions of `pyodide-build` build wheels with this tag, which is identical to the Emscripten version tag. The tag is named as such because we break the Emscripten ABI twice per year (at most).
- some miscellaneous documentation-only changes + changes to the grammar/punctuation

Closes #147

Issue #147, opened 2024-12-12
micropip 0.7 will break piplite (Pyodide 0.27)


juntyr:
micropip 0.7 switches the index_urls and fetch_kwargs parameters around in the query_package function, which pipeline patches-over. Once pyodide-kernel updates to a Pyodide with this new micropip (probably 0.27), the patch needs to be fixed. Otherwise, piplite fails to load any packages.

jtpio:
Thanks @juntyr for the heads-up!

Once a pre-release is available, it would be nice to open a PR to start looking into these changes.

agriyakhetarpal:
> Thanks @juntyr for the heads-up!
> 
> Once a pre-release is available, it would be nice to open a PR to start looking into these changes.

Thanks for filing this issue, @juntyr, since, as mentioned in the issue linked above, we didn't know that these APIs were being used. @jtpio, a pre-release dated to the end of August is available at https://github.com/pyodide/pyodide/releases/tag/0.27.0a2, but it probably did not include the `micropip` 0.7.1 version bump via pyodide/pyodide#5179 (I cannot check fully, as I'm typing from my phone ATM).

I saw that https://github.com/jupyterlite/pyodide-kernel/releases/tag/v0.5.0a1 was released early last week. Is there a chance a fix for this could be included with 0.5.0a2 or with 0.5.0 stable? We hope to start rolling Pyodide 0.27 out for Scientific Python projects' interactive documentation uses downstream as soon as we release this week, and this item would be a blocker (or one of them).

Merry Christmas!

juntyr:
I think the dependency on an internal API comes from the layering of JupyterLite’s packages over Pyodide’s packages. Now that micropip supports multiple index URLs, could JupyterLite build a static JSON index of its packages and pass that to micropip?

bollwyvl:
The current 0.5.x effort is mostly reacting to the JupyterLab/JupyterLite API changes for stability and performance.

When a pyodide release drops, we're generally fairly fast and thorough when supporting new upstream `pyodide` releases, but I wouldn't bet on any timelines. We'll support it, either only on the 0.5.x line, or on 0.4.x, depending on the release window and time permitting.

jtpio:
> We hope to start rolling Pyodide 0.27 out for Scientific Python projects' interactive documentation uses downstream as soon as we release this week,

[...]

PR #154, opened 2025-01-01
[0.4.x] Support Pyodide 0.27

bollwyvl:
## References
- addresses https://github.com/jupyterlite/pyodide-kernel/issues/147
- alternative to #153 (with manual cherry-picks)
  - targets `0.4.x` instead of `main` (just 

## Changes
- [x] bump `pyodide` in all the various places
- [x] handle API differences 
- [x] from #153
  - [x] invert order of `piplite.piplite._query_package`'s `index_urls` and `fetch_kwargs` 
  - [x] apply spelling/branding updates  
  - [x] add new accepted wheel `pyodide_*_wasm32.whl`
    - [x] kept `WASM_WHL` constant to avoid needless breakage
  - [x] update reference `piplite` CLI spec with support annotations
  - [ ] ???


agriyakhetarpal (approved):
Thanks!

> add new accepted wheel pyodide_*_wasm32.whl
>
> kept WASM_WHL constant to avoid needless breakage

xref to the issue about this in `micropip` here: pyodide/micropip#158. Should I put together a PR to resolve this in `micropip` first before this new ABI tag lands here (and subsequently, have it in a new `micropip` 0.8.1 release via a Pyodide 0.27.1 release)?

Either way, I think this PR looks good for Pyodide 0.27.0 and `micropip` 0.8.0 in specific, and the above can be taken into account with another [0.4.y] update after this one.

agriyakhetarpal:
Actually, never mind, that was fixed with https://github.com/pyodide/micropip/pull/159 which made it to the 0.7.2 release already, it is just that the issue was not closed (I'll mark it as closed now). All good, sorry for the confusion!

bollwyvl:
> never mind, that was fixed

Yeah, that's why I don't bother following `micropip` too tightly: the only version i ever look at is _what gets shipped to CDN_. 

It's... non-trivial to replace it from a `jupyterlite` perspective, as at least today, we must bootstrap a couple patched wheels, and `pyodide-lock` isn't fit for this purpose, as it won't generate a solve without a full runtime which has lead me to rather... [absurd lengths](https://pypi.org/project/jupyterlite-pyodide-lock) to get to something user-serviceable for the 80% case of "my noarch wheel and its deps."

I don't think we can do anything about the `@types/emscripten` vs `pyodide/types.d.ts` at this point, it was nice while it lasted, and every `as any` makes the baby Isaac Newton (maybe Leibniz) cry.

I am trying some more notebooks from jupyterlite's docs on RTD, and _may_ do some more abusive checks with those wheels, but otherwise, this is likely ready for more review and testing.

agriyakhetarpal:
Thanks for the additional context! We are planning to unvendor recipes with Pyodide 0.28 (it is our near term plan). Would it be worthwhile to unvendor `micropip` and treat it as a special package, given how extensively it is used here, so that handling releases is easier?

Yes, `pyodide-lock` also has a few issues, though I have not been involved with it much. But, I'm not sure if https://pypi.org/project/jupyterlite-pyodide-lock/ would also fully work (it's awesome if it does), because the last time I tried using a custom wheel with JupyterLite, I ran into something similar to https://github.com/jupyterlite/jupyterlite/issues/1502 (it has been a while, though).

bollwyvl:
Yeah, jlpl works, but only through a rather ridiculous contraption of browser automation, and a custom web server/proxy. 

Until such a time as we can do an accurate `micropip.freeze` solve in a "regular" CPython, it's the best I've got.

> unvendor recipes

Yeah, no skin in the game there. We'll continue to minimize user-space breakage, and support upstream features as best we can here.

> handling Pyodide releases

I wouldn't worry too much about us: that's why use all those types, have tests, and generally try to get our downstream releases out quickly, but _only_ after an actual upstream release. There are too many moving upstream parts for us to do much else.

jtpio (approved):
Thanks!

jtpio:
@meeseeksdev please backport to main

jtpio:
Looks like we still need a manual backport to `main`.

In the meantime, starting a new `0.4.7` release with this.

bollwyvl:
Cool, thanks for the release. Also merged the conda-forge PR, will see about getting the beta out a tick.

Issue #60, opened 2021-06-29
Configure Pyolite to have packages automatically installed


martinRenou:
### Problem

It is currently not possible to automatically install Python packages inside the pyolite kernel.

### Proposed Solution

It would be nice to be able to configure Pyolite to include some packages that would automatically be installed with `micropip`.

cc. @psychemedia who suggested it here https://github.com/jtpio/jupyterlite-demo/pull/7#issuecomment-870144941

### Additional context

It might be possible to have it once we use IPython for the execution (see related work https://github.com/jtpio/jupyterlite/pull/171), as it's possible to configure IPython itself to run code at the shell start.

bollwyvl:
With [pyodide-lock 0.1.0a4](https://github.com/pyodide/pyodide-lock/blob/v0.1.0a4/pyodide_lock/utils.py#L154), there is now enough API surface to do this in a semi-sane, less-hacky fashion.

The `pyodide-lock.json` approach _will_  allow for:
- participating in the lazily-loaded auto-import behavior

However, as a _lock_ (and not an _index_), it will not allow for packages that _might_ have multiple versions at runtime, depending on prior imports... and i'm not entirely sure `micropip`'s built-in patching features (even if further patched) will allow for it, so the fake-Warehouse-like API (or some follow on, such as the simple HTML or JSON file APIs) probably needs to stay, for now.

I have some related work which would benefit significantly from this, but am not yet up to where I can take advantage of it fully.

juntyr:
See https://github.com/jupyterlite/pyodide-kernel/pull/167#issuecomment-2655900778 for how to preload packages from the pyodide lockfile where the package list comes from a URL query parameter.

PR #158, opened 2025-01-10
Allow to install extra packages from extra indexes from jupyter-lite.json

Carreau:
This is in progress, but the idea is to be able to point to dev wheels on rtd or similar without having to make a custom pyidide distribution.

For example:

```
// jupyter-lite.json
{
  "jupyter-config-data": {
    "litePluginSettings": {
       "extraPackagesAndIndexes": [
          {
            "indexes": [
              "https://cors.carreau.workers.dev/scientific-python-nightly-wheels/simple",
              "https://pypi.org/simple"
            ],
            "package": "xarray"
          }
        ]
      }
    }
  }
```

Package could also be a simple url to a whl (I need to check how it works if it's a local URL), and indexes can be ommited.

WIP as I'm stuggling to have the typescript build process work.


Carreau:
cc @agriyakhetarpal

packages/pyodide-kernel/src/kernel.ts
```diff
@@ -394,5 +395,6 @@ export namespace PyodideKernel {
      * The Jupyterlite content manager
      */
     contentsManager: Contents.IManager;
+    extraPackagesAndIndexes: Array<{ package: string; indexes: string[] | null }>;
```
Carreau, line 398: make that `Array<{ package: string[]; ... }>;`

packages/pyodide-kernel/src/tokens.ts
```diff
@@ -90,6 +90,11 @@ export namespace IPyodideWorkerKernel {
      */
     mountDrive: boolean;
 
+    /**
+     * List of extra url/wheel paths to load, and extra associated indexes urls (if not a wheel url)
+     */
+    extraPackagesAndIndexes: Array<{ package: string; indexes: string[] | null }>;
```
Carreau, line 96: same here.

packages/pyodide-kernel/src/worker.ts
```diff
@@ -80,6 +85,28 @@ export class PyodideRemoteKernel {
...
+        try {
+          await this._pyodide.runPythonAsync(installCmd);
+          console.info(`Package ${pkg} Installed successfully`);
+        } catch (e) {
+          console.error('Error installing package', e);
+        }
+      }
+    } else {
+      console.info('no extra packages and indexes');
+    }
```
Carreau, line 109: Maybe move that to initKernel.

packages/pyodide-kernel/src/worker.ts
```diff
@@ -80,6 +85,28 @@ export class PyodideRemoteKernel {
...
 
     // get piplite early enough to impact pyodide-kernel dependencies
     await this._pyodide.runPythonAsync(`
```
Carreau, line 112: I think lower in initKernel the 

```
    for (const pkgName of toLoad) {
      if (!preloaded.includes(pkgName)) {
        scriptLines.push(`await piplite.install('${pkgName}', keep_going=True)`);
      }
    }
```

Could be modified to be a single call to `piplite.install(list)`, instead of repeated install calls.

packages/pyodide-kernel/src/worker.ts
Carreau: And use piplite instead of micropip ?

packages/pyodide-kernel/src/worker.ts
```diff
@@ -80,6 +85,28 @@ export class PyodideRemoteKernel {
...
+
+    if (extraPackagesAndIndexes.length > 0) {
+      // note that here pkg can be a package name or a wheel url
+      for (let { package: pkg, indexes } of extraPackagesAndIndexes) {
+        let installCmd: string;
+        if (indexes === null) {
+          installCmd = `import micropip\nawait micropip.install('${pkg}', keep_going=True)`;
+        } else {
+          installCmd = `import micropip\nawait micropip.install('${pkg}', index_urls=${JSON.stringify(indexes)}, keep_going=True)`;
+        }
```
Carreau, line 98: In both cases, if pkg ends with whl, prepend baseurl, 
and from testing, check the index urls do not end in `/`, otherwise it redirects and there are CORS issues.

Carreau:
`https://cors.carreau.workers.dev/scientific-python-nightly-wheels/simple` should not be used in production of course.

Carreau:
and cc @gabalafou

bollwyvl:
Hmmmm, while I take no issue with the indexes (indeed, maybe a `pipliteInstallDefaultOptions` would allow for overriding _all_ of the non-`requirements` [`micropip.install` kwargs](https://github.com/pyodide/micropip/blob/v0.8.0/micropip/install.py#L17-L23)), we've been trying not to add new sources of run-time unpredictability _before_ a user can make their first input and get useful feedback in the UI, which _really_ should be all the kernel is on the hook for... but yeah, this is the browser, and ain't nothing easy.

As of `0.4.7`, upstream's `pyodide-lock.json` is missing only `comm`, as `ipython` (and its 100k of vendored `tests`) is good to go. Pulling in a single wheel by URL might bring in many other, unpredictable wheels, which could (and it sounds like in this case, with `nightly`) _expect_ to be different between two different site visits, and very possible breaking when installed against other things, and stirred together with dependencies from PyPI.

So my feeling would be: we could add hooks to allow _another_ extension to load extra packages, run dynamic startup code, configure environment variables, write files on disk, make telemetry posts, etc. and all the other  things that folk have asked for, but _well_ after the core plumbing has loaded so that _when_ errors happen, those messages appear, with attribution, somewhere.

Carreau:
Sure that would be fine with me, what Im really hopping is to allow most of the Python Ecosystem to be able to get interactive documentation without having to rebuild a pyodide distribution. 

I'm fine if extra dependencies and breakage are up to the maintainers' using this configuration option. 

I'm also happy to move this to another place in the launching process.

Carreau:
crosslinking to #60

Carreau:
Would you agree if this were change to only look at locally available wheels ? 

Basically when using `jupyter lite build  --piplite-wheels`; I don't think the users should need an extra install step;

For me including wheels in a jupyterlite build should make them immediately available to the end user, or it feels weird. I'm ok if wheels needs to be referred to in two locations.

agriyakhetarpal:
> Basically when using jupyter lite build --piplite-wheels; I don't think the users should need an extra install step;
>
> For me including wheels in a jupyterlite build should make them immediately available to the end user, or it feels weird. I'm ok if wheels needs to be referred to in two locations.

I would agree with this, as this is one of the points that I've made in sympy/live#26 and it would be nicer to see that `micropip`/`piplite` consider pre-installing only if they find all wheels locally (say, if I were to ship SymPy and all its dependencies – mpmath, together) and look at the local index at a priority over Pyodide.

gabalafou (commented):
As promised, I gave this a look and it looks good to me. 

It seems that your comment about getting TypeScript to work was based on a previous commit and that you've resolved the issues with the TypeScript compiler, is that correct?

packages/pyodide-kernel/src/worker.ts
```diff
@@ -102,6 +102,25 @@ export class PyodideRemoteKernel {
...
+        let installCmd: string;
+        if (indexes === null) {
+          installCmd = `await piplite.install(${JSON.stringify(pkgs)}, keep_going=True)`;
+        } else {
+          installCmd = `await piplite.install(${JSON.stringify(pkgs)}, index_urls=${JSON.stringify(indexes)}, keep_going=True)`;
+        }
+        console.info('installing via cmd:', installCmd);
+        try {
+          await this._pyodide.runPythonAsync(installCmd);
+          console.info(`Package ${pkgs} Installed successfully`);
```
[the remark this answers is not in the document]
Carreau: well, but if you start two notebooks, you will get two install commands running concurrently...

Carreau:
> So my feeling would be: we could add hooks to allow _another_ extension to load extra packages, run dynamic startup code, configure environment variables, write files on disk, make telemetry posts, etc. and all the other things that folk have asked for, but _well_ after the core plumbing has loaded so that _when_ errors happen, those messages appear, with attribution, somewhere.

It seem to be that this asking way more than just having packages ready to use. Right now usually all the file and setup are ready, the only step missing is having pre-installed package. I understand the need to not wanting to download and extra URL, but there is already the --piplite-wheels so we only need to make them available at import.

Now if you really want a different extension fine, we can do tham but assuming I provide the necessary patches for `install(...)` to only look into the lock file and `all.json`, would that be accepted here ?

bollwyvl:
There is a big difference between blocking the kernel being available, with proper error reporting, on:

- _using `js.loadPackage` to install a known-good wheel, which comes from the same host as `pyodide.js`, and is guaranteed to have working dependencies_
- _use this random package, maybe without a pin, that depends on other packages, maybe without pins, downloaded from somewhere on the internet, that may have changed since the site was built_

There was a previously-rejected PR (don't see it here, maybe was before the SplitLite) that did the scut work to make that happen (discovering top-level names, etc.), by patching the as-loaded `pyode-lock.json`, but _still_ wasn't a solver, so could definitely get things _wrong_,  _and_ added extra dependencies and complexity. 

For those reasons, I've been pursuing upstream means to make this more predictable, such as:
- https://github.com/pyodide/micropip/pull/172
- https://github.com/pyodide/micropip/pull/181

...and downstream means to make this work in `pyodide-kernel`:

- https://github.com/deathbeds/jupyterlite-pyodide-lock

Maybe once _all_ of that works, every time, without npm/nodejs or a full browser at build time, we can consider making something that isn't too bad in this repo.

PR #161, opened 2025-01-17
use a list for piplite_urls

bollwyvl:
## references
- https://github.com/jupyterlite/jupyterlite/issues/1563
  - probably others, but needed to realize it was a `traitlets.config` issue

## changes
- [x] use `List` instead of `TypedTuple` for `piplite_urls`
- [x] test 0, 1, and 2 entries
- [x] handle duplicate wheels, e.g. `--piplite-urls . --piplite-urls ./some-package....whl`
  - [x] emits a `log.warning`, but doesn't fail

## user-facing changes
- site builders should be able to pass 0, 1, or _n_ wheels
  - if multiple wheels with exactly the same name are discovered, only one will be indexed (not sure which, but better than failing)
- site users should see no change

## backwards-incompatible changes
- i guess `piplite_urls` is a list now, not a tuple. but the tuple typing was wrong anyway, so likely nobody will notice

## future work
- we really need upstream to have better types. 
  - `doit` makes this somewhat tricky, as there is no community-maintained `types-doit`, and we really only care about a very small proportion of them
  - `jupyterlite_core` is probably easier, but there's still rather a lot going on

### DIFF
--- a/jupyterlite_pyodide_kernel/addons/piplite.py
+++ b/jupyterlite_pyodide_kernel/addons/piplite.py
@@ -6,7 +6,7 @@ import re
 import urllib.parse
 from hashlib import md5, sha256
 from pathlib import Path
-from typing import Tuple as _Tuple
+from typing import List as _List
 
 import doit.tools
 from jupyterlite_core.constants import (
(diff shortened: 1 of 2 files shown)


jtpio:
@meeseeksdev please backport to 0.5.x

PR #164, opened 2025-01-30
pre-install packages concurrently

Carreau:
I think that instead of doing N requests of await piplite.install, we can do one request with all the required packages.

This seem to only affect 6 packages when adding logging:

     - ssl
     - sqlite3
     - ipykernel
     - comm
     - pyodide_kernel
     - ipython

I'm assuming it won't make that much of a difference.

### DIFF
--- a/packages/pyodide-kernel/src/worker.ts
+++ b/packages/pyodide-kernel/src/worker.ts
@@ -97,10 +97,10 @@ export class PyodideRemoteKernel {
     const scriptLines: string[] = [];
 
     // use piplite for packages that weren't pre-loaded
-    for (const pkgName of toLoad) {
-      if (!preloaded.includes(pkgName)) {
-        scriptLines.push(`await piplite.install('${pkgName}', keep_going=True)`);
-      }
+    const packagesToInstall = toLoad.filter((pkgName) => !preloaded.includes(pkgName));
+    if (packagesToInstall.length > 0) {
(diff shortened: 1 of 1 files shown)


Carreau:
@bollwyvl I think wrote the original, so maybe has a reason to do so.

bollwyvl:
Order certainly _was_ important at some point: there are some [patches applied](https://github.com/jupyterlite/pyodide-kernel/blob/main/packages/pyodide-kernel/py/pyodide-kernel/pyodide_kernel/patches.py) where certain things had to happen between two other things. This list used to be longer, but looks like we got it down to setting `MPLBACKEND`. I am not sure of the impact of _not_ setting that before something else happening, and indeed, maybe that's all handled by `matplotlib_inline`.

On the main, the biggest win is to not play: ideally we'd further drive down the number of things handled here before thinking that parallelism is going to make anything better.

I think the addition of `ssl` was a poor choice: it's quite large, and (by observation) never needed for bootstrapping the package manager/kernel, which is all that should ever be _required_ in this list. The failure mode in user content is gross, though, and I didn't see a viable way to sniff it, as some libraries make decisions on some things it provides.

`sqlite` is also quite chunky... and I don't think IPython's history stuff probably even really does anything useful in this context. Maybe we could patch that out.

And then, of course, as discussed a number of times, is the `IPython.tests`. The pyodide distribution will unvendor some tests into a sibling zip, and likely should be applied in this case.

PR #167, opened 2025-02-07
Allow customizing the lockFileURL with a query parameter

juntyr:
Custom `pyodide-lock.json` URLs provide an easy way to customize the Pyodide kernel. At the moment, you can only statically change it in the Pyodide config. With this PR, it would be possible to provide the URL as a `?pyodide-kernel-lock-file-url=...` query parameter. This should be a sufficient MVP to allow experimenting with custom kernels, from which we could later design a more user-friendly interface.

### DIFF
--- a/packages/pyodide-kernel/src/worker.ts
+++ b/packages/pyodide-kernel/src/worker.ts
@@ -52,10 +52,18 @@ export class PyodideRemoteKernel {
       importScripts(pyodideUrl);
       loadPyodide = (self as any).loadPyodide;
     }
-    this._pyodide = await loadPyodide({
+    const loadPyodideOptions = {
       indexURL: indexUrl,
       ...options.loadPyodideOptions,
-    });
+    };
+    // support loading the pyodide-lock.json file from a custom URL provided by
+    //  ?pyodide-kernel-lock-file-url=...
+    const urlParams = new URLSearchParams(window.location.search);
+    const lockFileURL = urlParams.get('pyodide-kernel-lock-file-url');
+    if (lockFileURL) {
+      loadPyodideOptions.lockFileURL = lockFileURL;
+    }
+    this._pyodide = await loadPyodide(loadPyodideOptions);
   }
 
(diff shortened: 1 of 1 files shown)


bollwyvl:
As mentioned on the issue about offering multiple kernelspecs: I don't see how how this would make the situation more predictable. Further, once very public API features like this get added, they get very hard to take away or change, as "it broke my WordPress integration that hot links to your docs site" or whatever.

As usual: the recommended approach to adding more features like this would be a separate extension or kernel.

juntyr:
I’d be happy with any approach that can provide dynamically selected Pyodide configs, right now I just need to change the lock file URL. Lock files have the advantage that they are fully resolved and my use case is about preventing dependency drift.

I’m not sure how an extension could help here without duplicating the entire Pyodide kernel. Are there some existing hooks for how another extension could change the launch params for a fresh kernel at runtime?

juntyr:
I published an extension, https://github.com/juntyr/a-jupyterlite-query-config, that allows you to enable overriding any (config-allowed) config with URL query params. To use,

requirements.txt for the JupyterLite build:
```requirements.txt
# JupyterLite: override config settings with URL query parameters
a-jupyterlite-query-config~=0.1.0
```

jupyter-lite.json:
```json
{
    "jupyter-lite-schema-version": 0,
    "jupyter-config-data": {
        "litePluginSettings": {
            "a-jupyterlite-query-config:plugin": {
                "overrides": {
                    "pyodideKernelLockFileURL": "litePluginSettings.@jupyterlite/pyodide-kernel-extension:kernel.loadPyodideOptions.lockFileURL",
                    "pyodideKernelPackages": "litePluginSettings.@jupyterlite/pyodide-kernel-extension:kernel.loadPyodideOptions.packages"
                }
            }
        }
    }
}
```

to allow overriding the lock file URL with `?pyodideKernelLockFileURL="URL"` and the package preload list with `?pyodideKernelLockFileURL=["PACKAGE"]`

PR #169, opened 2025-02-13
Support passing of index URLs to piplite (CLI, runtime configuration, and requirements files)

### LINKED ISSUE #166: Add a `--index_urls` CLI flag for pip-install magics
### Problem

`piplite` supports passing a string or a list of strings of index URLs

https://github.com/jupyterlite/pyodide-kernel/blob/36aedb97d38cf283df8c3b0d65d7d6b8e575f0ab/packages/pyodide-kernel/py/piplite/piplite/piplite.py#L134

but it does not yet provide a CLI flag for them, so it is currently not possible to write commands `%pip install numpy --index-urls https://anaconda.org/scientific-python-nightly-wheels/simple`, as extra indices are/will be ignored and NumPy 2.0.2 from the Pyodide distribution is installed instead.

### Proposed Solution

Extend https://github.com/jupyterlite/pyodide-kernel/blob/36aedb97d38cf283df8c3b0d65d7d6b8e575f0ab/packages/pyodide-kernel/py/piplite/piplite/cli.py to have a simple `--index-urls` option that is passed to `piplite.py`, at first. It could be extended to `--index-url` or `--extra-index-url` options later if and when such a change is brought upstream in `micropip`.

### Additional context

xref: https://github.com/pyodide/pyodide/issues/4898#issuecomment-2637731077, as I've been meaning to test out the rollout of CORS headers for the Anaconda.org PyPI-like index through `piplite`, but I found I was unable to do so, while `micropip` works.

agriyakhetarpal:
## Description                                                                                                                                                                                           
                                                                                                                                                                                                      
Users can now specify alternate package indices when installing packages through `piplite` via three interfaces: the CLI (`-i`/`--index-url`), requirements files, and JupyterLite-site-wide configuration in           
`jupyter-lite.json`.                                                                                                                                                                                    

Closes #166

## Changes made

These are mostly along the lines of https://github.com/jupyterlite/pyodide-kernel/issues/166#issuecomment-2639857708.

### In the Python layer (`piplite`)

- Added `-i`/`--index-url` flag to the piplite CLI, forwarded as index_urls to `micropip.install`
- Updated the requirements file parsing so that `--index-url` directives are extracted from `-r requirements.txt` files and applied to all packages in that file, matching the behaviour of `pip`
(https://stackoverflow.com/a/2477610). However, the CLI flag takes priority if both are provided.
- Consolidated all JavaScript ➡️ Python runtime configuration into a single `_PIPLITE_DEFAULT_INSTALL_ARGS` dict (replacing the previous separate `_PIPLITE_URLS` and `_PIPLITE_DISABLE_PYPI` module-level variables),
which now also carries `index_urls` from `pipliteInstallDefaultOptions`

### In the TypeScript/configuration layer

- Defined a new `IPipliteInstallOptions` interface for representation of `piplite` install options
- Extended `kernel.v0.schema.json` to accept `pipliteInstallDefaultOptions`, with `index_urls` accepting a single URL string or a list
- The way this is wired, the extension reads `pipliteInstallDefaultOptions` from `jupyter-lite.json` ➡️ `PyodideKernel` ➡️ `initRemoteOptions` ➡️ `worker` ➡️ sets `_PIPLITE_DEFAULT_INSTALL_ARGS` keys via
`runPythonAsync` on kernel start

<hr>

The intended behaviour in `piplite` will look like the following:

1. via the CLI

`%pip install numpy -i https://pypi.myindex.com/simple`

2. via a requirements file

#### requirements.txt

```text
--index-url https://pypi.myindex.com/simple
numpy
scipy
```

and then `%pip install -r requirements.txt` processes the index at the top line.

3. via `jupyter-lite.json` (i.e., a site-wide default)

```json
{
  "jupyter-config-data": {
    "litePluginSettings": {
      "@jupyterlite/pyodide-kernel-extension:kernel": {
        "pipliteInstallDefaultOptions": {
          "index_urls": ["https://pypi.myindex.com/simple"]
        }
      }
    }
  }
}
```

Note that this still requires PyPI to be used effectively, since some packages the kernel relies on may not be present in `myindex` and need to be installed from PyPI. So, in real-life situations, we're most likely to see both `myindex` and PyPI configured.


agriyakhetarpal:
The fact that it's tricky to get a development version installed for me does not help either 😅 It took me a few tries but testing in CI is probably more convenient.

agriyakhetarpal:
`jlpm lint:py:check` and `jlpm:lint:py` are passing for me locally – I don't immediately see what the CI is complaining about.

agriyakhetarpal:
> [lite-badge](https://jupyterlite-pyodide-kernel--169.org.readthedocs.build/en/169/_static/lab/index.html) 👈 Try it on ReadTheDocs

Trying to test this with the Anaconda index does not seem to work as the CORS headers have been blocked for some reason again... I'll ping at https://github.com/pyodide/pyodide/issues/4898.

agriyakhetarpal:
A bit more testing reveals that this works well. Here are a few points about things I noticed, and some questions:

1. If I try to install NumPy (which should give me version 2.3.0dev0 at the time of writing), it gets us version 2.0.2, and NumPy gets loaded from the defaults channel. This is because NumPy is being fetched from

    https://github.com/jupyterlite/pyodide-kernel/blob/99f58ab4e8dbc2a8be67cdd7cd5096650aed349f/examples/jupyter-lite.json#L7-L9

	at the moment – it should work well outside of this PR, as it works 

2. I am successfully able to install packages listed in a `requirements.txt` file. This is working with both cases:
    a. When I use `%pip install -r requirements.txt -i https://myindexurl.com/simple`
    b. When I use %pip install -r requirements.txt` and the index URL is listed in the topmost line of `requirements.txt`

3. It is a bit of a pain to install packages with just `--index-url`. For example, `%pip install sympy -i "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/"` does not work, as `mpmath` being its dependency exists on PyPI and not with this index. I have to call `%pip install mpmath` before installing a nightly version of SymPy. This is going to be a bit of a bother for other packages as well, especially those that are consumers of packages hosted on limited indices such as Anaconda.org SPNW in this case. Would it make sense to also support `--extra-index-url`? One downside would be that it would also make the users' browsers' environments prone to dependency confusion attacks, which we would definitely not want. My use case with the functionality I'm trying to offer through this PR is that I should be able to write `%pip install mypackage -i https://myindexurl.com/simple`, where it would install `mypackage` through the passed index URL, but fetch the dependencies for `mypackage` from PyPI (this is possible maybe with `disablePyPIFallback` set to `False`?) or Pyodide's jsDelivr CDN (with the latter being preferred, of course) if they don't exist on the passed index URL. This is so that I won't need to write multiple pip-install magics for multiple dependency installations. In some sense, I would like `--index-url` to behave like `--extra-index-url`, which would be different from `pip`'s behaviour, but still making sense for a platform like JupyterLite (which could be documented). Another solution for this could be to deviate from `pip`'s behaviour of supporting just a single index URL with a requirements file, and supporting multiple URLs instead. Here is what a sample `requirements.txt` file with multiple URLs could look like:

	##### `requirements.txt`

	```text
    mypackage1
    mypackage2
    sympy
    mypackage3
    -r another_set_of_requirements.txt
    -i https://myindex1.com/simple # an index URL, only for the packages listed below up to another index URL if there is one
    scipy
	-i https://myindex2.com/simple # another index, only for numpy in this case
    numpy
    ```

	But I am not too keen on such an interface, as it is a bit of extra effort to implement and maintain. The current solution is not a big pain, it is manageable to some extent.

5. I haven't yet tried out the `jupyter-lite.json` and `pipliteInstallDefaultOptions` pathway I added yet (which is the only reason I haven't marked this fully as ready for review yet). I shall try to do so later in the day. Feedback in the meantime is still welcome!

6. What would be the ideal way to go about adding some tests for this? We don't want to query the indices unnecessarily as a part of a test suite in CI. In `micropip`, we have another test suite that is triggered with "real" (non-mocked) indices, which we run through through a manual workflow triggered by a `workflow_dispatch:` hook.

packages/pyodide-kernel-extension/schema/kernel.v0.schema.json
```diff
@@ -25,6 +25,21 @@
       "default": [],
       "format": "uri"
     },
+    "pipliteInstallDefaultOptions": {
+      "type": "object",
+      "description": "Default options to pass to piplite.install",
+      "default": {},
+      "properties": {
+        "indexUrls": {
```
bollwyvl, line 33: I would reckon this should be `index_urls`, and then wherever possible, the parent object handed around as a blob of JSON-compatible kwargs.

packages/pyodide-kernel/src/worker.ts
```diff
@@ -81,12 +86,20 @@ export class PyodideRemoteKernel {
     `);
     }
 
+    const pythonConfig = [
+      'import piplite.piplite',
+      `piplite.piplite._PIPLITE_DISABLE_PYPI = ${disablePyPIFallback ? 'True' : 'False'}`,
+      `piplite.piplite._PIPLITE_URLS = ${JSON.stringify(pipliteUrls)}`,
```
bollwyvl, line 92: this would then become `piplite.piplite._PIPLITE_DEFAULT_INSTALL_ARGS`

agriyakhetarpal:
The integration tests are now timing out. I'm not sure if this is due to a change I made... reverting and trying again

agriyakhetarpal:
I seem to have broken the installation of packages from an in-line index URL inside a `requirements.txt` file while addressing your review, @bollwyvl. Would it make sense to split adding a `--index-url` flag and the rest of your prior asks in #166 into separate PRs, or should I proceed to debug within the same PR?

My idea is that if we are to support multiple index URLs and multiple requirements files, each of those files could possibly use a non-default index URL, i.e., $n$ requirements files passed to the `pip`-install command could have $<= n$ index URLs for them (if the URLs are included within the files, that is – otherwise if an index URL is passed via the CLI, it would take priority). Handling this is becoming a bit messy, and at the same time, this is a rare scenario in real-life use cases.

agriyakhetarpal:
Never mind: actually, a local test to see `pip`'s behaviour reveals that it considers only the last index URL in the requirements file, i.e., it searches for packages in the last index URL it sees. Passing `--index-url` in the command with multiple requirements files containing an index of their own disregards them, as only the CLI one has any effect.

This should make things a bit more simpler.

Pro tip: never combine the act of installing packages from separate indices and requirements files, and invoke a `pip`-install separately for each requirements file if any of them has an inline index URL. This is a mess that I hope no one has to encounter, and one we don't have to support in too convoluted a manner 🤠

agriyakhetarpal:
I think this PR is ready to garner some more feedback.

About the `pipliteInstallDefaultOptions` setting I mentioned in point 4 of https://github.com/jupyterlite/pyodide-kernel/pull/169#issuecomment-2658194228, it isn't working yet, but I also have some second thoughts about implementing it. Having an index URL in place as a setting would constrain the user to look only through said index URL, which isn't a good solution as many dependencies might not be available. Is there a better interface in mind, or maybe I missed something?

https://github.com/jupyterlite/pyodide-kernel/issues/166#issuecomment-2639857708 mentioned adding it as a site-level configurable, but in this case, this acts more like `--index-url` and not `--extra-index-url` (the latter, again, is susceptible to dependency confusion as I mentioned somewhere above). There is a way to add multiple indices through [`micropip.set_index_urls()`](https://micropip.pyodide.org/en/latest/project/api.html#micropip.set_index_urls) and we can call that with `piplite`, but maybe `micropip.install()` should have an `extra_index_urls` argument similar to `micropip.install(index_urls="<...>")`?

I feel we should limit my scope with this PR, but I am happy to implement what sounds best either upstream or here.

bollwyvl:
> Having an index URL in place as a setting would constrain the user to look only through said index URL, 

Yes. That's what site-level configurations _should_ do. We're answering a lot of different use cases, from the _wide open playground, hosted on a dumb public CDN_ to _an exam setting with very locked down CORS and a hot API token and icky telemtry_. 

This is why we also provide the ability to disable PyPI fallback altogether. While `micropip.install` might break on a given day, we'd expect `%pip install` to work... as mentioned elsewhere, we'd really prefer _not_ to be dependent on _any_ dependency resolution at kernel startup, much less known-to-be-changing upstreams. A PR at one point proposed hot-patching the JS package index with entries from e.g. `--piplite-wheels` (including built-ins like `pyodide-kernel`, etc.) but rejected as the pyodide repo (i mean index (i mean lock)) was too much of a moving target.

> which isn't a good solution as many dependencies might not be available. Is there a better interface in mind, or maybe I missed something?

Yes, for the time being, a site administrator would be on the hook to configure the `index_urls` to be the appropriate list. If they don't want to allow any upstreams:
```
{
  "index_urls": [
    "https://example.edu/pypi/simple"
  ]
}
```

Or if they intend to have just a few local packages:

```
{
  "index_urls": [
    "https://example.edu/pypi/simple",
    "https://pypi.org/simple",
  ]
}
```

We use the `extra_*` pattern in many places (both here and in `jupyterlite-core`), but... 

> `extra_index_urls`

... `micropip.install` should support that. 

From a `pyodide-kernel` perspective, complete PyPI indexes sufficient to bootstrap the kernel (prior to `ipython` and friends being in `pyodide-lock.json`) are non-trivial to set up. Again, most desirable would be to attempt to mimic `pip install` here, offering both `--index-url` and `--extra-index-url`...  but as that upstream behavior basically _encourages_ typosquatting, reverse the behavior (`--extra-index-urls` would be _prepended_ to the list). But again, we really don't want to take on any more complexity than necessary.

packages/pyodide-kernel/py/piplite/piplite/cli.py
```diff
@@ -37,6 +37,11 @@ async def install(
 
 REQ_FILE_SPEC = r"^(?P<flag>-r|--requirements)\s*=?\s*(?P<path_ref>.+)$"
 
+# Matches --index-url or -i directives (with optional = separator and optional quotes)
+INDEX_URL_SPEC = (
+    r'^(--index-url|-i)\s*=?\s*(?:"([^"]*)"|\047([^\047]*)\047|([^\s]*))\s*$'
```
bollwyvl: similar to the above about `uri-reference`, but also, trying to write "real" URL matches in regexen is one of those things that will almost certainly break in some case.

bollwyvl (changes requested):
Probably still some more work/discussion needed.

PR #171, opened 2025-02-26
pyodide 0.27.3

bollwyvl:
## changes
- [x] update to `pyodide 0.27.3` in the usual places (schema, constant, `package.json`)

## notes
- most relevant changelog item
  > Since 0.27.1, Pyodide has been broken in iOS because iOS ships broken wasm-gc support. Pyodide feature detects whether the runtime supports wasm-gc and uses it if it is present. Unfortunately, iOS passes the feature detection but wasm-gc doesn't work as expected.


bollwyvl:
Seems fine on RTD. I don't have an iOS device with which to test.

jtpio (approved):
Thanks!

Issue #88, opened 2023-11-01
How to access Pyodide APIs to customize init


billti:
I've read through the docs, a bunch of issues, and played around with a few of the templates (the GitHub starter and "jupyter lite" commands), but can't figure out how to do this, so maybe an area for docs improvement.

How do you access the Pyodide APIs to configure the pyodide kernel before user environments (REPL, Notebooks) are made available in JupyterLite?

In this particular case, a package I want to use is only shipped as native binaries and doesn't have a pyodide compatible build available, but can be easily provided by a JavaScript shim and registered via `pyodide.registerJsModule` (see https://pyodide.org/en/stable/usage/faq.html#how-do-i-create-custom-python-packages-from-javascript ), so I want to run this before-hand in the kernel at startup. How should this be done?

jtpio:
> I can have a look if this looks interesting.

Thanks!

I guess this could indeed be experimented with. And made opt-in, so site administrators would have to explicitly choose to enable such feature.

jtpio:
I moved the issue to the https://github.com/jupyterlite/pyodide-kernel repo since this seems to be more specific to that kernel.

Alternatively, it may be possible to implement such behavior as a more generic JupyterLab extension, that would work with any kernel and also with JupyterLab (just an idea, haven't checked more into the details to see if it's feasible).

PR #172, opened 2025-02-27
Add bootstrapCode config option

juntyr:
The new `bootstrapCode` extension config option allows JupyterLite hosts to run custom Python code before the kernel starts. This should help downstream projects with the commonly wanted feature of "running code before the first cell".

Related to #88

### DIFF
--- a/packages/pyodide-kernel-extension/schema/kernel.v0.schema.json
+++ b/packages/pyodide-kernel-extension/schema/kernel.v0.schema.json
@@ -37,6 +37,11 @@
           }
         }
       }
+    },
+    "bootstrapCode": {
+      "description": "Python code that is executed by the Pyodide kernel during initialization. Beware that throwing an exception during bootstrap will stop the kernel from starting.",
+      "type": "string",
+      "default": ""
     }
   }
 }
--- a/packages/pyodide-kernel-extension/src/index.ts
+++ b/packages/pyodide-kernel-extension/src/index.ts
@@ -58,6 +58,7 @@ const kernel: JupyterLiteServerPlugin<void> = {
     const pipliteUrls = rawPipUrls.map((pipUrl: string) => URLExt.parse(pipUrl).href);
     const disablePyPIFallback = !!config.disablePyPIFallback;
     const loadPyodideOptions = config.loadPyodideOptions || {};
+    const bootstrapCode = config.bootstrapCode || '';
 
     for (const [key, value] of Object.entries(loadPyodideOptions)) {
       if (key.endsWith('URL') && typeof value === 'string') {
@@ -98,6 +99,7 @@ const kernel: JupyterLiteServerPlugin<void> = {
           disablePyPIFallback,
           mountDrive,
           loadPyodideOptions,
+          bootstrapCode,
           contentsManager,
         });
       },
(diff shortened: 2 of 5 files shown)


juntyr:
ping @jtpio

jtpio:
Thanks @juntyr.

Having it implemented via such option sounds like a fine approach.

Also when there is something to help better see the logs (https://github.com/jupyterlite/jupyterlite/issues/1498), it will be easier for users to troubleshoot issues if something goes wrong with this extra `bootstrapCode`.

bollwyvl:
Again, this is a feature that also does not exist in JupyterLab, and for which there are user space extensions.

For "pre-installing" packages, there are _also_ other approaches such as [jupyterlite-pyodide-lock](https://github.com/deathbeds/jupyterlite-pyodide-lock), which put everything known packages into `pyodide-lock.json` such that they will be imported automatically.

juntyr:
> Again, this is a feature that also does not exist in JupyterLab, and for which there are user space extensions.

Could you please point towards one? If there’s an existing extension that can do that I’d be very happy to use it and close this PR.

> For "pre-installing" packages, there are _also_ other approaches such as [jupyterlite-pyodide-lock](https://github.com/deathbeds/jupyterlite-pyodide-lock), which put everything known packages into `pyodide-lock.json` such that they will be imported automatically.

That just preloads them, but doesn’t run them. What this PR tries to provide is a way to run custom code during bootstrap, which is not yet possible otherwise (as far as I know). At the moment, I have to use some very hacky patches to IPython to get some code to run after the interpreter is setup but before the user gains control. This is very valuable for applying behind-the-scenes magic that the user doesn’t need to see.

bollwyvl:
Here is a discourse discussion with more context (and related links at the bottom):

https://discourse.jupyter.org/t/setting-initialization-cells-for-jupyterlab/13345

juntyr:
I found that this can be done using IPython startup scripts or extensions. See https://stackoverflow.com/questions/32438235/ipython-notebook-default-initialization-python-code, https://modelpredict.com/how-to-write-reliable-scripts-ipython/, https://stackoverflow.com/questions/48803339/modifying-jupyter-notebook-in-init-code, and https://emscripten.org/docs/porting/files/packaging_files.html#modifying-file-locations-in-the-virtual-file-system

jtpio:
@juntyr would you still like to document your approach, for example in https://jupyterlite.readthedocs.io/en/stable/howto/index.html#configuring-the-pyodide-kernel? So it can be useful to other people too.

juntyr:
Right now the approach requires controlling the Pyodide build, as I need to make emscripten embed the IPython config files into the filesystem. The rest is already documented by https://modelpredict.com/how-to-write-reliable-scripts-ipython/, I just copied its code.

So in the end we have:
1. I now have an assets folder in my Pyodide fork here https://github.com/climet-eu/pyodide/tree/058717a2b1311b9efbd3166e5c5ff35a12d75bf6/assets, which contains files that should be included in the filesystem
2. I modified Pyodide's Makefile to embed the folder here https://github.com/climet-eu/pyodide/blob/058717a2b1311b9efbd3166e5c5ff35a12d75bf6/Makefile#L129
3. The asset folder contains an ipython_config.py file that adds the extensions here https://github.com/climet-eu/pyodide/blob/058717a2b1311b9efbd3166e5c5ff35a12d75bf6/assets/home/pyodide/.ipython/profile_default/ipython_config.py
4. The extension, which can be a file in the same assets folder, exposes the following interface

```python
def load_ipython_extension(ip):
    # do stuff here
```

PR #175, opened 2025-03-19
Update to JupyterLite 0.6.0a3

jtpio:
Update to https://github.com/jupyterlite/jupyterlite/releases/tag/v0.6.0a3.

Follow-up to https://github.com/jupyterlite/jupyterlite/pull/1590.

PR #176, opened 2025-03-19
Bump to JupyterLite 0.6.0a4

jtpio:
Follow-up to https://github.com/jupyterlite/pyodide-kernel/pull/175

Also bump to https://github.com/jupyterlite/jupyterlite/releases/tag/v0.6.0a4

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8109,7 +8109,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"source-map-js@npm:^1.0.2, source-map-js@npm:^1.2.1":
+"source-map-js@npm:^1.2.1":
   version: 1.2.1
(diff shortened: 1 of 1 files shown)


bollwyvl:
As mentioned on the other issue: we really ought to have this under lint/test: at least _some_ upstream version in  `package.json` and `pyproject.toml` should always be in lockstep.

jtpio:
Right, or even a workflow to do these updates automatically more easily.

I've started using a workflow / update script for updating Notebook to the latest JupyterLab releases, and this has already helped reduce maintenance fatigue quite a bit: https://github.com/jupyter/notebook/blob/main/.github/workflows/upgrade-jupyterlab-dependencies.yml

PR #182, opened 2025-04-10
Remove dependency on `IBroadcastChannelWrapper`, use `browsingContextId`

jtpio:
Depends on https://github.com/jupyterlite/jupyterlite/pull/1621

- [x] Remove the dependency on `IBroadcastChannelWrapper`
- [x] Use `browsingContextId` when instantiating a new `DriveFS`
- [x] Update to JupyterLite 0.6.0a8
- [x] Remove tmp patches

PR #183, opened 2025-04-25
Support stdin via service worker

ianthomas23:
This adds support for `stdin` requests using JupyterLite's service worker. The python commands that use this are `input` and `getpass.getpass` which currently work in the pyodide kernel but need to be `await`ed.  With this PR the `await` will no longer be required, provided JupyterLite's service worker is available. If the service worker is not available the functions will not work. It is part of the work of jupyterlite/jupyterlite#275 to support synchronous `stdin` across all JupyterLite kernels.

The changes here are not sufficient on their own, a JupyterLite PR is also required which I intend to submit next week. I would like the changes here to be reviewed and merged and a pre-release made so that the JupyterLite PR can run CI against this.

### DIFF
--- a/packages/pyodide-kernel/src/worker.ts
+++ b/packages/pyodide-kernel/src/worker.ts
@@ -5,6 +5,8 @@ import type Pyodide from 'pyodide';
 
 import type { DriveFS } from '@jupyterlite/contents';
 
+import { URLExt } from '@jupyterlab/coreutils';
+
 import { KernelMessage } from '@jupyterlab/services';
 
 import type { IPyodideWorkerKernel } from './tokens';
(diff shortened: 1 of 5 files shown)


packages/pyodide-kernel/src/worker.ts
```diff
@@ -449,48 +451,72 @@ export class PyodideRemoteKernel {
    * @param content The incoming message with the reply
    */
   async inputReply(content: any, parent: any) {
-    await this.setup(parent);
-
-    this._resolveInputReply(content);
+    // Should never be called as input_reply messages are returned via service worker
```
ianthomas23, line 454: We could explicitly throw an exception here in case this function is ever called in the future.

jtpio:
> I would like the changes here to be reviewed and merged and a pre-release made so that the JupyterLite PR can run CI against this.

Note this is not strictly required to test this PR with a dev build of JupyterLite.

For previous PRs that involved the Service Worker (https://github.com/jupyterlite/pyodide-kernel/pull/182), we used the built wheel for `jupyterlite-pyodie-kernel` which is uploaded to ReadTheDocs.

For #182 it was https://jupyterlite-pyodide-kernel--182.org.readthedocs.build/en/182/_static/jupyterlite_pyodide_kernel-0.6.0a4-py3-none-any.whl.

For this PR it will be: https://jupyterlite-pyodide-kernel--183.org.readthedocs.build/en/183/_static/jupyterlite_pyodide_kernel-0.6.0a5-py3-none-any.whl

That way it's easier to iterate faster and avoids making many releases for new fixes.

ianthomas23:
I have added a `SharedArrayBuffer` implementation for stdin, similar to that for `xeus` kernels.

jtpio:
cc @martinRenou in case you would also like to review this one as the companion of https://github.com/jupyterlite/jupyterlite/pull/1640

martinRenou (approved):
🚀

jtpio (approved):
Thanks!

PR #201, opened 2025-07-18
WIP, publish error tracebacks on iopub

tomjakubowski:
This ensures that when an Output ipywidget is used as a context manager, errors raised inside the context are rendered in the Output widget.

```
out = Output()
out

with out:
    raise RuntimeError("oh no! an error occurred")
```

However, it does not suppress the error in the original cell, which is to say the error is rendered both in the original cell and in the Output widget.  This behavior is in contrast to Jupyter Lab and Notebook, which do suppress the error in the original cell.  Fixing this is a TODO.

This patch also handles the case where an ipywidgets handler, without the `@out.capture()` decorator, raises an exception.  Such exceptions are now seen in the log console.

TODOs:
- [ ] Suppress error in the original cell

### DIFF
--- a/packages/pyodide-kernel/py/pyodide-kernel/pyodide_kernel/display.py
+++ b/packages/pyodide-kernel/py/pyodide-kernel/pyodide_kernel/display.py
@@ -58,6 +58,7 @@ class LiteDisplayHook(DisplayHook):
     def __init__(self, *args, **kwargs):
         super(LiteDisplayHook, self).__init__(*args, **kwargs)
         self.publish_execution_result = None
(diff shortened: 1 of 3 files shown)


tomjakubowski:
When I say this patch fails to suppress the error, I mean this. Compare the cell outputs in Jupyter Notebook with ipykernel: 

Screenshot 2025-07-17 at 11 16 59 PM

and the cell outputs in Jupyter Lite, with this patch to pyodide-kernel:

Screenshot 2025-07-17 at 11 17 25 PM

PR #203, opened 2025-08-05
Update to jupyterlite-core 0.7.x, pyodide 0.28.1, micropip 0.10.1

bollwyvl:
## references

- fixes #199 
- based on (and alternative to) #202

## changes
- [x] use `jupyterlite-core 0.7.0a1` in the usual places
- [x] use `jupyterlab 4.5.0` in the usual places
- [x] use `pyodide 0.28.1` in the usual places
- [x] update reference from `pyodide.PyodideInterface` to `pyodide.PyodideAPI`
- [x] update `piplite` for `micropip 0.10.1`
  - [x] add CLI args `--reinstall` and `--constraints`
  - [ ] ???

## backwards incompatible changes
- bumps to `python 3.13` and `emscripten 4`
  - as discussed in a couple places, this seems big enough to warrant starting an `0.7.x` line, even if `0.6,x` (which would stay on `pyodide 0.27`) continues to get updates for the next `jupyterlite-core`

## alternatives
- see #202

## outstanding questions
- as there is no particular timeline for `lab 4.5.0`, should this actually be matrixed further:

| pr | version | lab | jupyterlite-core | pyodide |
|-|-|-|-|-|
| | `0.6.x`  | `4.4.x` | `0.6.x` | `0.27.x` |
| #202 | `0.7.x` | `4.4.x` | `0.6.x` | `0.28.x` |
| this | `0.8.x` |  `4.5.x` | `0.7.x` | `0.28.x` |

PR #204, opened 2025-08-06
Fix reinstall arg

jtpio:
Follow-up to #203


bollwyvl:
Thanks for catching this :P

I never got around to checking all the upstream notebooks, will do so now.

bollwyvl:
Had a look through everything:
- it's possible to get some `setTime` errors when doing things fast
  - maybe happens on `0.6.3`
- `--force-reinstall` "works" (it updated the files on disk)
  - but then `importlib.reload` _doesn't_ work
  - so maybe instead we pull the feature?

Issue #205, opened 2025-08-07
Investigate more Pyodide 0.28 / micropip changes


jtpio:
Had a look through everything:
- it's possible to get some `setTime` errors when doing things fast
  - maybe happens on `0.6.3`
- `--force-reinstall` "works" (it updated the files on disk)
  - but then `importlib.reload` _doesn't_ work
  - so maybe instead we pull the feature?

_Originally posted by @bollwyvl in https://github.com/jupyterlite/pyodide-kernel/issues/204#issuecomment-3161793147_

PR #208, opened 2025-08-26
Update Pyodide to v0.28.2

agriyakhetarpal:
This PR is the same as #202, but against the `main` branch.

TODO items:

- [x] Update Pyodide version everywhere
- [ ] Bump micropip
	- [ ] Add new wrapper APIs

See also: #205


agriyakhetarpal:
Oh, actually, it looks like the new `micropip` changes have been added in #203 already. In that case, is there more that we need to do here?

agriyakhetarpal:
I'd be happy to share a few ideas on #205 when I have a moment to sit down and compile them, if that helps. :)

PR #210, opened 2025-09-01
Trying patch approach

martinRenou:
It looks to me that the patching approach does not seem to work, trying to confirm that with a dummy patch that should make the kernel crash on any `open`


martinRenou:
Closing in favor of #211

PR #242, opened 2025-11-20
use eslint consistent-type-imports

bollwyvl:
## references
- continues #240
- see also upstream
  - https://github.com/jupyterlite/jupyterlite/pull/1786
  - https://github.com/jupyterlab/jupyterlab/pull/18129

## changes
- [x] add `@typescript-eslint/consistent-type-imports`
- [x] run fixers

## notes
- I didn't see any further reductions in worker sizes (still 33/180kb), but perhaps when the above upstreams all land, some more good might happen

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,6 +119,13 @@
       "@typescript-eslint/no-explicit-any": "off",
       "@typescript-eslint/no-namespace": "off",
       "@typescript-eslint/no-use-before-define": "off",
+      "@typescript-eslint/consistent-type-imports": [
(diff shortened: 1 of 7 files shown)


martinRenou:
> I didn't see any further reductions in worker sizes (still 33/180kb), but perhaps when the above upstreams all land, some more good might happen

Yeah, "some good might happen" is really the outcome of this I think. Although I've had some packages where that `import type` was crucial so we'll see, maybe it will have a good impact.

PR #247, opened 2025-12-19
Update to JupyterLite 0.8.0a0

jtpio:
Update to https://github.com/jupyterlite/jupyterlite/releases/tag/v0.8.0a0


jtpio:
Probably something we can investigate separately. For now this PR still builds the worker separately.

Will follow-up with a first pre-release.

PR #248, opened 2026-01-02
update to pyodide 0.29.1

bollwyvl:
Changes:
- [x] update to `pyodide 0.29.1` in various places
- [x] update `requires-python` for vendored wheels (e.g. `piplite`)
- [x] add test for `requires-python`

Motivation:

It looks like we can _almost_ use `uv pip compile --python-platform=wasm32-pyodide2024` to get solves without a browser, but all of the metadata has to be more accurate than what `micropip` has been letting us get away with.

Alternatives:
- [ ] set the `requires-python` to something _much_ lower (e.g. `3.10`)
- [ ] remove `requires-python` altogether


bollwyvl:
Thank you!

PR #275, opened 2026-03-29
[0.8] Add pyodide-lock.json customization with optional pyodide-lock and uv dependencies

bollwyvl:
## references
- forward-port of #269 
- https://github.com/jupyterlite/jupyterlite/pull/1916
  - demo on readthedocs:  [](https://jupyterlite--1916.org.readthedocs.build/en/1916/_static/lab/index.html)

## changes

> vs #269

- [x] split out `lock.py` 
- [x] add `-r requirements.txt` and `-g ../pyproject.toml:group` parsing for `PyodideLockAddon.specs` and `.constraints`
- [x] add writing out `PyodideLockAddon.lite_constraints` for future solves 
  - > updating `pyproject.toml#dependency-groups` is too hard 
  - [x] allow `widgetsnbextension` and `jupyterlab-widgets` to be included in lock/constrants to influence successive solves
- [ ] ???


jtpio:
Sorry for the late reply here.

Will take a look today, but probably the best would be to just merge and release in a 0.8.0 pre-release so it's easier to test and we can fix things later during the pre-release cycle.

bollwyvl:
@jtpio updated the demo pr over on `jupyterlite`... would love to see this in before we start ramping up for the beta cycle, mucking about with `jupyter-builder`, etc.

jtpio (approved):
Thanks, let's try!

bollwyvl:
Thanks for the release! Working on the conda-forge PR, noted that something's up with the schema check test, and needs and `admin-requests` PR, but otherwise running smoothly.

I'll circle back around on the demo PR, but might be kind of a busy this weekend.

PR #290, opened 2026-05-11
Mark jupyterlite packages as shared

martinRenou:


martinRenou:
As discussed with @jtpio this has no impact since the web worker cannot share packages with the main thread

bollwyvl:
This might be worth revisiting with the latest `jupyter-builder 1.0` release which depends on `rspack 2.x`; however, it might be that some deeper config is needed in either `jupyterlite-core` and/or, probably better, `jupyter-builder` itself, if there is some low-coordination way to get this properly.

bollwyvl:
(see also https://github.com/jupyterlab/jupyterlab/issues/10033)

Issue #292, opened 2026-06-03
PyodideKernel.run() leaks SyntaxError/TabError from input transformation as a raw kernel traceback (Python 3.12+)


JimWallace:
## Summary

When a code cell contains a `SyntaxError` that CPython's tokenizer raises *during
tokenization* — most commonly `TabError: inconsistent use of tabs and spaces in
indentation` — it is not surfaced as a normal error cell. The exception escapes
`PyodideKernel.run()` unhandled, and the user sees a raw internal traceback through
`pyodide_kernel/litetransform.py` → `IPython/core/inputtransformer2.py:make_tokens_by_line`,
with no indication that they simply mixed tabs and spaces.

## Environment

- `jupyterlite-pyodide-kernel` 0.7.2
- Pyodide 0.28 (Python 3.13), IPython 9.0.2
- Also confirmed against IPython **9.14.0** (latest) — this is **not** version-specific.

## Minimal reproduction

Run this cell. The `elif`/`else` lines (and their `return`s) are indented with a TAB
character while the rest of the body uses spaces:

```python
def f(a, b):
    if a:
        return 1
	elif b:        # <- this line begins with a TAB
		return 2   # <- this line begins with a TAB
```

**Actual:** a raw traceback ending in `make_tokens_by_line` →
`generate_tokens_catch_errors`.

**Expected:** a clean error cell:

```
  Cell In[1], line 4
    	elif b:
           ^
TabError: inconsistent use of tabs and spaces in indentation
```

## Root cause

`PyodideKernel.run()` calls the input transforms *outside* any `try/except`:

```python
code = await self.lite_transform_manager.transform_cell(code)
exec_code = self.interpreter.transform_cell(code)
```

On Python 3.12+ the C tokenizer raises `TabError`/`IndentationError` (a `SyntaxError`
subclass) eagerly during tokenization (see python/cpython#105238). IPython's
`generate_tokens_catch_errors` only does `except tokenize.TokenError`, so it cannot
catch a `SyntaxError`; both `make_tokens_by_line` (called directly by
`litetransform.do_one_token_transform`) and `InteractiveShell.transform_cell`
propagate it.

IPython's own `InteractiveShell.run_cell` guards `transform_cell` in a `try/except`
and renders the error via `showtraceback()`/`showsyntaxerror()`. `run()` reimplements
the cell loop and omits that guard.

Note: this is **not** fixable by upgrading IPython (confirmed against 9.14.0), because
[...]

PR #293, opened 2026-06-03
Surface SyntaxError from input transformation instead of crashing the…

JimWallace:
… kernel

PyodideKernel.run() called transform_cell() outside any try/except, so a SyntaxError raised during input transformation — most commonly TabError on Python 3.12+, which the tokenizer raises eagerly — escaped run() unhandled and surfaced as a raw kernel traceback instead of a normal error cell.

Guard the transform like InteractiveShell.run_cell does and route failures through showtraceback() (which dispatches SyntaxError to showsyntaxerror), reusing the existing _last_traceback reporting.

https://claude.ai/code/session_017qs8Cz1XgFG4oBPqNPqLom

### DIFF
--- a/packages/pyodide-kernel/py/pyodide-kernel/pyodide_kernel/kernel.py
+++ b/packages/pyodide-kernel/py/pyodide-kernel/pyodide_kernel/kernel.py
@@ -100,24 +100,37 @@ class PyodideKernel(LoggingConfigurable):
 
     async def run(self, code):
         self.interpreter._last_traceback = None
-        # apply pyodide-specific changes that need to occur before interpreting
-        code = await self.lite_transform_manager.transform_cell(code)
-        exec_code = self.interpreter.transform_cell(code)
 
         results = {}
 
+        # apply pyodide-specific changes that need to occur before interpreting
         try:
-            await _load_packages_from_imports(exec_code)
+            code = await self.lite_transform_manager.transform_cell(code)
+            exec_code = self.interpreter.transform_cell(code)
         except Exception:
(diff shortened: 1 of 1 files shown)


JimWallace:
Closes #292 

## Summary
On Python 3.12+, a `SyntaxError` that CPython's tokenizer raises during tokenization — most commonly `TabError: inconsistent use of tabs and spaces in indentation` — escaped `PyodideKernel.run()` unhandled, so the user saw a raw internal traceback through `pyodide_kernel/litetransform.py` → `IPython/core/inputtransformer2.py:make_tokens_by_line` instead of a normal error cell.

## Root cause
`PyodideKernel.run()` called the input transforms outside any `try/except`. On Python 3.12+ the C tokenizer raises `TabError`/`IndentationError` (a `SyntaxError` subclass) eagerly during tokenization (python/cpython#105238). IPython's `generate_tokens_catch_errors` only does `except tokenize.TokenError`, so it cannot catch a `SyntaxError`; both `make_tokens_by_line` and `InteractiveShell.transform_cell` propagate it. IPython's own `InteractiveShell.run_cell` guards `transform_cell` in a `try/except`; `run()` reimplemented the cell loop and omitted that guard. Not fixable by upgrading IPython (confirmed against 9.14.0).

## Fix
Wrap the two `transform_cell` calls in a `try/except` that routes failures through `showtraceback()` (which dispatches `SyntaxError` → `showsyntaxerror`), nesting the existing `_load_packages_from_imports` block under `else:`. Reuses the method's existing `_last_traceback` → `results["status"] = "error"` reporting.

PR #294, opened 2026-06-10
Update to Pyodide v314.0.0

agriyakhetarpal:
A mechanical update in all the usual places, and also dropping `ssl` and `sqlite3`. The main change is the bundling configuration, as we are now an ES6 module: https://blog.pyodide.org/posts/314-release/#pyodide-is-now-a-native-es-module, which also led to a refactor.

Towards https://github.com/jupyterlite/jupyterlite/issues/1844
Towards https://github.com/pyodide/pyodide/issues/6075


agriyakhetarpal:
Looks like there are some test failures. I'll take a closer look later today. Thanks!

bollwyvl:
Before digging too deeply through test fails, it looks like we may have to do... rather a lot to adapt to [`pyodide.mjs`](https://github.com/pyodide/pyodide/blob/314.0.0/docs/project/changelog.md?plain=1#L49):

```js
client.js:149 Uncaught (in promise) Failed to execute 'importScripts' on 'WorkerGlobalScope': 
The script at 'https://cdn.jsdelivr.net/pyodide/v314.0.0/full/pyodide.js' failed to load.
```

If the module situation can't be resolved easily, we may have to fully re-work how things work at build time and/or runtime. Statically compiling the `pyodide.*.mjs` into the two workers would be pretty bad, fetching them at build time might be feasible, but needs to be at least smoke tested in a number of server/browser combinations.

agriyakhetarpal:
So I think I got a bit closer to getting it to work, but I soon hit

`Uncaught (in promise) Classic web workers are not supported`

when instantiating the kernel.

We probably need to fix `@jupyterlab/builder`'s `rspack` config so that `new Worker(new URL(..., import.meta.url), {type:'module'})` can emit an ES module worker chunk. Or can we override the webpack config for extensions? I will have to do some more readings about this.

agriyakhetarpal:
I decided to throw Opus 4.6 at the problem (I was wanting to try out the new Fable 5, but it doesn't seem to be working for me 🤷🏻‍♂️), and I think I have gotten to something promising enough, i.e., I kind of agree with what it did – while I have some idea about web bundlers, I am by no means an expert or know deeply about their context within the Jupyter ecosystem. Basically, what it did was rework the Webpack config such that Rspack's `WorkerPlugin` cannot downgrade `{type: 'module'}` workers to classic workers, which we refuse with Pyodide 314.0. So when Rspack cannot bundle the workers and `@jupyter/builder` cannot override the base rule, our rules merge after the base rule. So it emits prebundled workers as assets and not chunks.

I previously tried `output.module: true` and that caused some module chunk-ID collision where the worker was loading the wrong bundle, and that didn't really seem right to me. Looks like some other tests are failing now, but I tested the kernel in a deployment locally and everything seems to work correctly for me. Edit: and I tested on RTD, works well there too.

But I'm sure this approach might not be entirely correct and will have limitations that I cannot express given that I haven't followed along the worker-based peculiarities of the kernel code.

packages/pyodide-kernel/src/worker.ts
```diff
@@ -43,17 +43,33 @@ export abstract class PyodideRemoteKernel {
 
   protected async initRuntime(options: IPyodideWorkerKernel.IOptions): Promise<void> {
     const { pyodideUrl, indexUrl } = options;
-    let loadPyodide: typeof Pyodide.loadPyodide;
-    if (pyodideUrl.endsWith('.mjs')) {
-      // note: this does not work at all in firefox
-      const pyodideModule: typeof Pyodide = await import(
-        /* webpackIgnore: true */ pyodideUrl
+
+    // Pyodide 314.0.0 and later only ships an ES module runtime, and it
```
bollwyvl: I don't think it's really worth the full paragraph of comments; something short, with a link to (preferrably upstream) docs/changelog/migration guide is sufficient. 

Considering the comments (even if shortened) would primarily be of interest for downstream wrapper kernel authors, consider refactoring some of these up to methods with docstrings so that they end up in e.g. IDE completion.

packages/pyodide-kernel/src/worker.ts
```diff
@@ -43,17 +43,33 @@ export abstract class PyodideRemoteKernel {
...
+        `Loading ${resolvedPyodideUrl} instead of the configured ${pyodideUrl}. ` +
+          'Pyodide 314.0.0 and later can only be loaded as an ES module.',
       );
-      loadPyodide = pyodideModule.loadPyodide;
-    } else {
-      importScripts(pyodideUrl);
-      loadPyodide = (self as any).loadPyodide;
     }
+
+    // Load the ES module entry point with a plain dynamic import. The
```
bollwyvl: Same comment about comment length/position; this may be worth being fully extracted to a separate utility file.

README.md
```diff
@@ -77,6 +77,7 @@ yet work in a full, `jupyter_server`-hosted client such as JupyterLab or Noteboo
 |      `>=0.6.0,<=0.7.0`       | `0.27.*`  | `3.12.*` |   `3.1.58`   |                |            |
 |      `>=0.7.0,<=0.8.0`       | `0.29.*`  | `3.13.*` |   `4.0.9`    |                |            |
 |      `>=0.8.0,<=0.9.0`       | `0.29.*`  | `3.13.*` |   `4.0.9`    | `>=0.1.3,<0.2` | `>=0.9.27` |
+|      `>=0.9.0,<=0.10.0`      |  `314.*`  | `3.14.*` |   `5.0.3`    | `>=0.1.3,<0.2` | `>=0.9.27` |
```
bollwyvl: this should be an update to the previous `0.8` line.

jupyterlite_pyodide_kernel/constants.py
```diff
@@ -24,16 +24,20 @@
 
 #: directory name and filenames for the Pyodide distribution
 PYODIDE = "pyodide"
+#: the legacy classic-script entry point. Pyodide still ships it, but it
+#: cannot be loaded from the kernel's module workers, so it is no longer used.
 PYODIDE_JS = "pyodide.js"
```
bollwyvl: please remove this entirely; a downstream is better off failing to import the now-broken old variable.

agriyakhetarpal:
Thanks for the suggestions @bollwyvl - updated! I linked back to this PR and our release blog post as these two seem like the most useful pieces to refer to. And I refactored into the Pyodide loading into a new `loader.ts` file too.

packages/pyodide-kernel/src/worker.ts
```diff
@@ -43,17 +45,8 @@ export abstract class PyodideRemoteKernel {
...
-      const pyodideModule: typeof Pyodide = await import(
-        /* webpackIgnore: true */ pyodideUrl
-      );
-      loadPyodide = pyodideModule.loadPyodide;
-    } else {
-      importScripts(pyodideUrl);
-      loadPyodide = (self as any).loadPyodide;
-    }
+    const pyodideModule = await importPyodideModule(pyodideUrl);
+    const { loadPyodide } = pyodideModule;
```
bollwyvl:
```suggestion
    const { loadPyodide } = await importPyodideModule(pyodideUrl);
```

packages/pyodide-kernel/src/worker.ts
```diff
@@ -120,14 +113,7 @@ ${e.stack}`;
...
 
-    const toLoad = [
-      'sqlite3',
-      'ipykernel',
-      'comm',
-      'pyodide_kernel',
-      'jedi',
-      'ipython',
-    ];
+    const toLoad = ['ipykernel', 'comm', 'pyodide_kernel', 'jedi', 'ipython'];
```
bollwyvl: Since we're changing this let's align with normalized hyphenated names, so `pyodide-kernel`.

agriyakhetarpal: Done!

bollwyvl:
Opened https://github.com/jupyterlite/jupyterlite/pull/1973 to see how it plays with the full stable of demo packages, especially `ipympl` (spoiler from local: not yet), against the wheels from _this_ docs build.  [_That_ docs build](https://jupyterlite--1973.org.readthedocs.build/en/1973/_static) exercises more of the potential gotchas...

packages/pyodide-kernel/src/loader.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Pyodide 314.0.0 and later is ESM-only and must be loaded from its
+ * `pyodide.mjs` entry point, see
+ * https://blog.pyodide.org/posts/314-release/. The `import()` call is
+ * wrapped in `new Function` to keep it out of bundler static analysis,
+ * which would otherwise rewrite it into a module lookup that cannot
+ * resolve a runtime URL. See
+ * https://github.com/jupyterlite/pyodide-kernel/pull/294 for details.
+ */
+export async function importPyodideModule(url: string): Promise<typeof Pyodide> {
```
bollwyvl: This can probably be generic over any module type, and doesn't really have anything to do with pyodide; while this kernel doesn't need to hot load anything else, a downstream very well could.

bollwyvl (approved):

agriyakhetarpal:
Thanks @bollwyvl – really appreciate the guidance on improving my TS! I think we can merge, as I understand from your comment on https://github.com/jupyterlite/jupyterlite/pull/1973 that it's now mostly about updating the notebooks, and the kernel is working fine as per our needs here.

jtpio:
Thanks @agriyakhetarpal and @bollwyvl!

@agriyakhetarpal if you would like to have this released, please feel free to make a new `v0.8.0b1` release, maybe it would be useful for testing before the final release? It should be a matter of using the (default) `next` version spec in the release workflow.

bollwyvl:
There is the dangling concern of `%matplotlib`, and the more general case of `entry_points` for other stacks like `nbconvert`; this might be something `pyodide-lock.json` could handle. `pyodide-build`'s upstream conda recipe format supports this for generating portable `console_scripts`, but in this case we'd want `matplotlib.backend` and all the fixings. Not going to happen anytime soon.

